PR #24, opened 2024-08-28
Proofset data, Create, Delete

### LINKED ISSUE #5: Proof set Create and Delete
Implement the state structures of the proof set logical array, their creation and deletion
1. Proof set Create: initialization of new proof set structure.  Including init of root id array, and sum tree index structure.  SetID generation and tracking
2. Proof set Delete: deletion of proof set data structures.  Ownership check.

ZenGround0:
Closes #5

### DIFF
--- a/contracts/src/PDPService.sol
+++ b/contracts/src/PDPService.sol
@@ -72,6 +72,9 @@ contract PDPService is Ownable {
 
     // Removes a proof set. Must be called by the contract owner.   
     function deleteProofSet(uint256 setId) public {
+        if (setId >= proofSetDeleted.length) {
+            revert("proof set id out of bounds");
+        }
         if (proofSetDeleted[setId]) {
             revert("Proof set already deleted");
         }
--- a/contracts/src/PDPService.sol
+++ b/contracts/src/PDPService.sol
@@ -6,12 +6,103 @@ import "@openzeppelin/contracts/access/Ownable.sol";
 contract PDPService is Ownable {
     // Types 
 
+    // TODO: there is more to think about here.
+    // existing libraries (https://github.com/filecoin-project/filecoin-solidity/blob/master/contracts/v0.8/types/CommonTypes.sol#L91)
+    // don't give us any significant functionality so it makes sense to redeclare.
+    // There will be performance benefits both for storage and memory ops to using bytes32 but we should wait 
+    // on measurment before making that decision.  We could potentially use bytes32 hash and bytes8 for the prefix.
+    struct Cid {
+        bytes data;
+    }
+
     // State fields
+
+    /*
+    A proof set is the metadata required for tracking data for proof of possession.
+    It maintains a list of CIDs of data to be proven and metadata needed to 
+    add and remove data to the set and prove possession efficiently.
+
+    ** logical structure of the proof set**
+    /* 
+    struct ProofSet {
+        Cid[] roots; 
+        uint256[] sizes;
+        uint256[] sumTree;
+        uint256 size;
+        bool deleted;
+    }
+    ** PDP service contract tracks many possible proof sets **
+    []ProofSet proofsets
+
+    To implement this logical structure in the solidity data model we have
+    two arrays tracking the singleton fields and three two dimensional arrays
+    tracking the growing data of the proof set.  The first index is the proof set id
+    and the second index is the index of the data in the array.
+
+    Invariant: rootCids.length == rootSizes.length == sumTreeSizes.length == proofSetSize.length
+    */
+
     uint256 public challengeFinality;
+    // TODO: a single mapping with a pair key i.e. (uint256, uint256 )
+    // might be more efficient as it will half the hashing needed for access and set
+    mapping(uint256 => Cid[]) rootCids; 
+    mapping(uint256 => uint256[]) rootSizes;
+    mapping(uint256 => uint256[]) sumTreeSizes;
+    bool[] public proofSetDeleted;
+    uint256[] public proofSetSize;
 
     // Methods
     constructor(uint256 _challengeFinality) Ownable(msg.sender) {
         challengeFinality = _challengeFinality;
+    }
+
+    // A proof set is created empty, with no roots. Creation yields a proof set ID 
+    // for referring to the proof set later.
+    function createProofSet() public returns (uint256) {
+        require(owner() == msg.sender, "Only the owner can create proof sets");
+        uint256 setId = proofSetSize.length;
+        proofSetSize.push(0);
+        proofSetDeleted.push(false);
+        rootCids[setId].push();
+        rootSizes[setId].push();
+        sumTreeSizes[setId].push();
+
+        return setId;
+    }
+
+    // Removes a proof set. Must be called by the contract owner.   
+    function deleteProofSet(uint256 setId) public {
+        if (proofSetDeleted[setId]) {
+            revert("Proof set already deleted");
+        }
+        require(owner() == msg.sender, "Only the owner can delete proof sets");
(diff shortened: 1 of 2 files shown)


anorth:
You could add a view method here to return the list of roots or perhaps even just the count if you want to defer that API design, and then test it for empty and deleted (if it remains) sets.

contracts/src/PDPService.sol
```diff
@@ -6,12 +6,106 @@ import "@openzeppelin/contracts/access/Ownable.sol";
 contract PDPService is Ownable {
```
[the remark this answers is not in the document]
ZenGround0: Since we're moving to singleton I'll move the ownership and documentation to proof set level

contracts/src/PDPService.sol
```diff
@@ -6,12 +6,106 @@ import "@openzeppelin/contracts/access/Ownable.sol";
...
+
+    Invariant: rootCids.length == rootSizes.length == sumTreeSizes.length == proofSetSize.length
+    */
+
     uint256 public challengeFinality;
+    // TODO: a single mapping with a pair key i.e. (uint256, uint256 )
+    // might be more efficient as it will half the hashing needed for access and set
+    mapping(uint256 => Cid[]) rootCids; 
+    mapping(uint256 => uint256[]) rootSizes;
+    mapping(uint256 => uint256[]) sumTreeSizes;
```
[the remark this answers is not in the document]
ZenGround0: Great points I was mistakenly under different assumptions about how dynamic arrays use slots.  

Since we're moving to singleton I prefer more performance isolation so will keep two mappings for now.

ZenGround0:
> I would generally lean towards having no public state variables and writing view functions for accessing that data. My understanding is that there is negligible, if any, gas cost difference. Solidity generates a view function under the hood anyway. Are there other considerations that led you to expose the deletion and size state directly?

I don't understand the benefit to keeping private with a view function vs public.  I don't mind changing but is there a concrete reason to?

ZenGround0:
Ok I'm seeing that the reason is [for encapsulation upon inheritance](https://forum.openzeppelin.com/t/why-are-state-variables-private-rather-than-public-in-openzeppelin-contracts/4387/3).  This is pretty weak but good enough for me.

contracts/src/PDPService.sol
```diff
@@ -6,12 +6,106 @@ import "@openzeppelin/contracts/access/Ownable.sol";
...
+    // A proof set is created empty, with no roots. Creation yields a proof set ID 
+    // for referring to the proof set later.
+    function createProofSet() public returns (uint256) {
+        require(owner() == msg.sender, "Only the owner can create proof sets");
+        uint256 setId = proofSetSize.length;
+        proofSetSize.push(0);
+        proofSetDeleted.push(false);
+        rootCids[setId].push();
+        rootSizes[setId].push();
+        sumTreeSizes[setId].push();
```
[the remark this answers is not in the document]
ZenGround0: I think this is just me learning how to solidity, gone now that we're using mapping mapping

anorth:
> Ok I'm seeing that the reason is [for encapsulation upon inheritance](https://forum.openzeppelin.com/t/why-are-state-variables-private-rather-than-public-in-openzeppelin-contracts/4387/3). This is pretty weak but good enough for me.

I agree the reason is weak, but enough to tip the default that way.

contracts/src/PDPService.sol
```diff
@@ -3,15 +3,151 @@ pragma solidity ^0.8.13;
...
+
+    // Network epoch delay between last proof of possession and next 
+    // randomness sampling for challenge generation
+    uint256 challengeFinality;
+   
+    // TODO PERF: https://github.com/FILCAT/pdp/issues/16#issuecomment-2329838769
+    uint64 nextProofSetId;
+    mapping(uint256 => mapping(uint256 => Cid)) rootCids; 
+    mapping(uint256 => mapping(uint256 => uint256)) rootSizes;
+    mapping(uint256 => mapping(uint256 => uint256)) sumTreeSizes;
```
anorth, line 49: Specifically for the sum tree, maybe an array in the inner collection makes sense? It'll always be dense, right?

contracts/src/PDPService.sol
```diff
@@ -3,15 +3,151 @@ pragma solidity ^0.8.13;
...
+    function getProofSetOwner(uint256 setId) public view returns (address) {
+        require(setId < nextProofSetId, "Proof set ID out of bounds");
+        return proofSetOwner[setId];
+    }
+
+    // Returns the root CID for a given proof set and root ID
+    function getRootCid(uint256 setId, uint256 rootId) public view returns (Cid memory) {
+        require(setId < nextProofSetId, "Proof set ID out of bounds");
+        require(rootId < nextRootId[setId], "Root ID out of bounds");
+        return rootCids[setId][rootId];
```
anorth, line 93: Here and in all the lookups above, this returns a default value (empty CID?) if the key is absent. This could be fine semantics, though might deserve some thought and documentation. We'll need to be careful of places where zero is an allowed value. For example, should we set nextProofSetID to 1 at construction, and similarly start root IDs at 1 so that zero can distinguish "no such proof set"?

If we do go with returning zero values for undefined mappings, there is not much point requiring the proof set ID or root ID to be within bounds. Is there any point distinguishing between not-yet-created vs deleted proof sets.

Ok to defer this, but it might affect multiple methods.

contracts/test/PDPService.t.sol
```diff
@@ -4,16 +4,58 @@ pragma solidity ^0.8.13;
...
+        vm.expectRevert("proof set id out of bounds");
+        pdpService.deleteProofSet(0);
+    }
+
+    function testDeleteAlreadyDeletedProofSetFails() public {
+        uint256 setId = pdpService.createProofSet();
+        pdpService.deleteProofSet(setId);
+        vm.expectRevert("Only the owner can delete proof sets");
+        pdpService.deleteProofSet(setId);
+    }
```
anorth, line 54: Add some tests for the view methods on a deleted proofset, if only as documentation.

anorth (approved):

contracts/src/PDPService.sol
ZenGround0: As far as I can tell one big downside with no IPLD is that we can't delete a whole map in constant time so we're stuck with all the leftover root IDs after "deleting" a proof set.  This is going to give us some weird results without some care.

ZenGround0: Yeah with more research it looks like the only way to actually delete state is to iterate all map keys which is possible but so expensive that no-one would do it.

If we want to support logical proof set remove we need to handle queries that 
1) overrun the proof set index
2) overrun the root id index
3) hit deleted proof sets
4) hit deleted root ids 

I'd like to do this consistently.  The two possibilities are revert or return zero value.  I vaguely prefer return zero value: ints 0, cids zero length byte arrays.

To do this correctly we need to check for all conditions as part of the view methods and explicitly return zero values in some cases.

ZenGround0: After a few minutes of implementing I'm now prefering revert since there are so many view functions and the zero return value makes different amounts of sense for different functions

ZenGround0: Ok i've settled on consistently reverting queries to deleted or not created proof sets with "Proof set not live" but returning zero values for the proof set arrays (size, sumtree, roots).  SumTree needs to do this for deleted roots and zero values are less confusing when you don't need to figure out whether its proof set or array that isn't live.

contracts/test/PDPService.t.sol
ZenGround0: Great idea

PR #26, opened 2024-09-05
Implemented Merkle proofs and verification

anorth:
The proof verification is adapted from OpenZeppelin libraries, with the hash changed to SHA256. The testing code includes building of trees from an array of roots, which we must replicate off-chain.

### DIFF
--- a/contracts/test/Proofs.t.sol
+++ b/contracts/test/Proofs.t.sol
@@ -1,0 +1,166 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.13;
+
+import {Test, console} from "forge-std/Test.sol";
+import {MerkleProof} from "../src/Proofs.sol";
+import {BitOps} from "../src/BitOps.sol";
+import {Hashes} from "../src/Proofs.sol";
+ 
+contract MerkleProofTest is Test {
+
+    function testVerifyEmptyProof() public view {
+        bytes32 root = sha256("hello");
+        bytes32[] memory proof = new bytes32[](0);
+        bool result = MerkleProof.verify(proof, root, root);
+        assertEq(result, true, "Verify should return true");
+    }
+
+    function testVerifyTreeTwoLeaves() public view {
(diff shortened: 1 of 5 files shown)


ZenGround0 (approved):
Suggestions for more documentation and testing.  But overall code looks great.

contracts/test/Proofs.t.sol
```diff
@@ -0,0 +1,158 @@
...
+        }
+        // Trim proof to the correct length, from proofIndex.
+        bytes32[] memory trimmedProof = new bytes32[](proofIndex);
+        for (uint256 i = 0; i < proofIndex; i++) {
+            trimmedProof[i] = proof[i];
+        }
+        return trimmedProof;
+    }
+
+    // Helper function to calculate the number of leading zeros in binary representation
```
ZenGround0, line 137: It turns out we're going to need this for the sum tree search function.  Could you create a library for low level binary operations and add this there?

ZenGround0: I'll then tweak [this function](https://github.com/FILCAT/pdp/pull/27/files#diff-712b2db21b090e745391534efb9ce9dfb70c3f729952e9433b627e264e3f83a3R212) rename to ctz, and also add to the library.

contracts/test/Proofs.t.sol
```diff
@@ -0,0 +1,158 @@
...
+        for (uint256 i = levels; i > 0; i--) {
+            bytes32[] memory currentLevel = tree[i];
+            uint256 nextLevelSize = (currentLevel.length + 1) / 2;
+            tree[i - 1] = new bytes32[](nextLevelSize);
+
+            for (uint256 j = 0; j < nextLevelSize; j++) {
+                if (2 * j + 1 < currentLevel.length) {
+                    tree[i - 1][j] = commutativeHash(currentLevel[2 * j], currentLevel[2 * j + 1]);
+                } else {
+                    tree[i - 1][j] = currentLevel[2 * j];
```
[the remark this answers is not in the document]
anorth: Will do

PR #28, opened 2024-09-09
Change Merkle proof scheme to verify leaf position

anorth:
We need to verify the position of a leaf and a protocol-determined offset, not any provider-chosen leaf.

### DIFF
--- a/contracts/src/Proofs.sol
+++ b/contracts/src/Proofs.sol
@@ -15,67 +15,77 @@ pragma solidity ^0.8.20;
 library MerkleProof {
     /**
      * Returns true if a `leaf` can be proved to be a part of a Merkle tree
-     * defined by `root`. For this, a `proof` must be provided, containing
-     * sibling hashes on the branch from the leaf to the root of the tree. Each
-     * pair of leaves and each pair of pre-images are assumed to be sorted.
+     * defined by `root` at `position`. For this, a `proof` must be provided, containing
+     * sibling hashes on the branch from the leaf to the root of the tree.
      *
      * This version handles proofs in memory.
      */
-    function verify(bytes32[] memory proof, bytes32 root, bytes32 leaf) internal view returns (bool) {
-        return processProof(proof, leaf) == root;
+    function verify(bytes32[] memory proof, bytes32 root, bytes32 leaf, uint256 position) internal view returns (bool) {
+        return processProofMemory(proof, leaf, position) == root;
     }
 
     /**
      * Returns the rebuilt hash obtained by traversing a Merkle tree up
-     * from `leaf` using `proof`. A `proof` is valid if and only if the rebuilt
-     * hash matches the root of the tree. When processing the proof, the pairs
-     * of leafs & pre-images are assumed to be sorted.
+     * from `leaf` at `position` using `proof`. A `proof` is valid if and only if the rebuilt
+     * hash matches the root of the tree.
      *
      * This version handles proofs in memory.
      */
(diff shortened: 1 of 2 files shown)


contracts/src/Proofs.sol
```diff
@@ -15,68 +15,78 @@ pragma solidity ^0.8.20;
...
-     * from `leaf` using `proof`. A `proof` is valid if and only if the rebuilt
-     * hash matches the root of the tree. When processing the proof, the pairs
-     * of leafs & pre-images are assumed to be sorted.
+     * from `leaf` at `position` using `proof`. A `proof` is valid if and only if the rebuilt
+     * hash matches the root of the tree.
      *
      * This version handles proofs in memory.
      */
-    function processProof(bytes32[] memory proof, bytes32 leaf) internal view returns (bytes32) {
+    function processProof(bytes32[] memory proof, bytes32 leaf, uint position) internal view returns (bytes32) {
```
ZenGround0, line 34: nit: maybe processProofMemory?

contracts/test/Proofs.t.sol
```diff
@@ -34,23 +34,33 @@ contract MerkleProofTest is Test {
...
+                    assertFalse(MerkleProof.verify(proof, root, leaves[i], j));
+                }
+            }
         }
     }
 
 
     function testVerifyTreesManyLeaves() public view {
-        for (uint256 width = 1; width < 100; width++) {
+        for (uint256 width = 4; width < 60; width++) {
```
ZenGround0, line 49: Why this change?

ZenGround0: Oh because you did 2 and 3 above? And 100 was overkill ?

ZenGround0 (approved):

contracts/test/Proofs.t.sol
anorth: Re `4`, yes. Re `100`, for some reason I don't fully understand, adding the "Ensure the proof is invalid for every other index" loop caused the test to consume more memory (not just quadratic compute) and fail. I tried a few tricks to release the memory, but without luck. Even 60 is probably way more than we need.

contracts/test/Proofs.t.sol
```diff
@@ -68,7 +78,7 @@ contract MerkleProofTest is Test {
     // Builds a merkle tree from an array of leaves.
     // The tree is an array of arrays of bytes32.
     // The last array is the leaves, and each prior array is the result of the commutative hash of pairs in the previous array.
-    // An unpaired element is simply copied to the next level.
+    // An unpaired element is paired with itself to create the value at the next level up.
```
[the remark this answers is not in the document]
anorth: My primary concern was simplicity and security. 

The old behaviour of promoting the unpaired element to the next layer (is that what you mean by identity hash?), thus shortening the proof, broke tracking of the position while verifying proofs. I did a little reading about how to handle odd nodes in this situation, and pairing with self (thus generating a new hash at each level, rather than it being predictable from just the leaf, or pairing with a constant) seemed recommended.

An alternatives we could look at:
- While verifying, given known width of the tree, track whether or not the last element is unpaired at each level and skip checking if our position is that last element. This would add some complexity to avoid one extra hash at each level (log in data size) while building the tree.

We should certainly seek some cryptographer review as we settle on things here.

contracts/test/Proofs.t.sol
ZenGround0: woah surprising

PR #29, opened 2024-09-11
Feat/proofset search

### LINKED ISSUE #8: SumTree Search
This is a prerequisite for implementing proof set `ProvePossession` 

Implement the [search function on the SumTree data structure](https://docs.google.com/document/d/16VtbihsKDEjzcCWJl0Rdz_M72itPzoM9GKZC1wgiB3Y/edit).  This will take in a challenge index into the proof sets byte array and return the index of the root id that contains the bytes referenced by the challenge index.

ZenGround0:
Closes #8

### DIFF
--- a/contracts/src/PDPService.sol
+++ b/contracts/src/PDPService.sol
@@ -5,7 +5,7 @@ import {BitOps} from "../src/BitOps.sol";
 
 contract PDPService {
     // Constants
-    uint256 public constant LEAF_SIZE = 256;
+    uint256 public constant LEAF_SIZE = 32;
 
     // Types
     // TODO PERF: https://github.com/FILCAT/pdp/issues/16#issuecomment-2329836995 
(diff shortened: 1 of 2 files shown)


contracts/src/PDPService.sol
```diff
@@ -174,12 +174,52 @@ contract PDPService {
         return rootId;
     }
 
-    // Removes a root from a proof set. Must be called by the contract owner.
+    // removeRoot removes a root from a proof set. Must be called by the contract owner.
```
anorth, line 177:
```suggestion
    // removeRoot removes a root from a proof set. Must be called by the proof set owner.
```

contracts/src/PDPService.sol
```diff
@@ -174,12 +174,52 @@ contract PDPService {
...
     function removeRoot(uint256 setId, uint256 rootId) public {
         require(proofSetOwner[setId] == msg.sender, "Only the owner can remove roots");
         // TODO: implement me
     }
 
+    // findRoot returns the root id for a given chunk index
```
anorth, line 183: We'll need to know the leaf/chunk offset within the tree committed by the root as well. I expect that this can be found easily enough as part of this method.

anorth (approved):

contracts/src/PDPService.sol
ZenGround0: Great point I'll add this into the return now

PR #32, opened 2024-09-17
Implement SHA254 hashing for the merkle tree

anorth:
This is only part of Filecoin piece CID compatibility, but should do the job for power-of-two data sizes.

### DIFF
--- a/contracts/src/Proofs.sol
+++ b/contracts/src/Proofs.sol
@@ -83,12 +83,12 @@ library MerkleProof {
 library Hashes {
     /** Order-dependent hash of pair of bytes32. */
     function orderedHash(bytes32 a, bytes32 b) internal view returns (bytes32) {
-        return _efficientSHA256(a, b);
+        return _efficientSHA254(a, b);
     }
 
 
-    /** Implementation of sha256(abi.encode(a, b)) that doesn't allocate or expand memory. */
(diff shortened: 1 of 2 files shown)


contracts/src/Proofs.sol
```diff
@@ -99,6 +99,12 @@ library Hashes {
             }
             
             value := mload(0x00)
+            // SHA254 hash for compatibility with Filecoin piece commitments.
+            // "The Sha254 functions are identical to Sha256 except that the last two bits of the Sha256 256-bit digest are zeroed out."
+            // The bytes of uint256 are arranged in big-endian order, MSB first in memory.
+            // The bits in each byte are arranged in little-endian order.
+            // Thus, the "last two bits" are the first two bits of the last byte.
```
ZenGround0, line 106: Thanks so much for the clear explanation

contracts/src/Proofs.sol
```diff
@@ -99,6 +99,12 @@ library Hashes {
...
+            value := and(value, 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF3F)
```
ZenGround0, line 107: This is fine but maybe consider extracting to a constant and moving comment along with it.  Im thinking this because we use this constant again in test

ZenGround0 (approved):

PR #33, opened 2024-09-20
Implement Filecoin piece-compatible padding of unbalanced Merkle trees

anorth:
Virtually zero-pads unbalanced trees out to a power of two with zero data.

I believe now that this will create an equivalent commitment to Filecoin given post-Fr32-padded data. Unfortunately I haven't yet found any test vectors for equivalence that don't also require doing Fr32 padding during the tree construction. Perhaps we can derive the Fr32-padded version of, e.g. `0xCCCCCCCC...` and then test against the vectors [here](https://github.com/filecoin-project/go-fil-commp-hashhash/blob/master/testdata/0xCC.txt).


contracts/src/Proofs.sol
```diff
@@ -78,6 +78,72 @@ library MerkleProof {
...
+     * A height of 1 returns the hash of two zero leaves.
+     * A height of n returns the hash of two nodes of height n-1.
+     * Height must be <= 50 (representing 2^50 leaves or 32EiB).
+     */
+    function zeroRoot(uint height) internal pure returns (bytes32) {
+        require(height <= 50, "Height must be <= 50");        
+        // These roots were generated by code in Proots.t.sol.
+        uint256[51] memory ZERO_ROOTS = [
+            0x0000000000000000000000000000000000000000000000000000000000000000,
+            0xf5a5fd42d16a20302798ef6ed309979b43003d2320d9f0e8ea9831a92759fb0b,
```
ZenGround0, line 94: Checks out with commp-hash-hash code I ported to go playground: https://go.dev/play/p/dICLMvpGIaA

ZenGround0 (approved):

PR #34, opened 2024-09-23
Batch find

ZenGround0:
This adds batching to the find operation by default to match the other state operations.

### DIFF
--- a/contracts/src/PDPService.sol
+++ b/contracts/src/PDPService.sol
@@ -196,15 +196,13 @@ contract PDPService {
         return delta;
     }
 
-    // findRoot returns the root id for a given leaf index and the leaf's offset within
-    // the rootId.
-    // 
-    // It does this by running a binary search over the logical array
-    // To do this efficiently we walk the sumtree.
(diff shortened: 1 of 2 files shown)


contracts/src/PDPService.sol
```diff
@@ -231,9 +229,24 @@ contract PDPService {
...
+    }
+
+    // findRoot returns the root id for a given leaf index and the leaf's offset within
+    // the rootId.
+    // 
+    // It does this by running a binary search over the logical array
+    // To do this efficiently we walk the sumtree.
+    function findRootIds(uint256 setId, uint256[] calldata leafIndexs) public view returns (RootIdAndOffset[] memory) { 
+        // The top of the sumtree is the largest power of 2 less than the number of roots
+        uint256 top = 256 - BitOps.clz(nextRootId[setId]);
```
ZenGround0, line 244: Maybe this is premature optimization -- the only reason this is here and not in findOneRootId is for the potential gas savings

anorth (approved):

PR #35, opened 2024-09-23
Devnet testing scripts

ZenGround0:
I've been using these scripts to do basic gas measurement and state inspection on a lotus devnet.  They seem useful enough to share.


tools/add.sh
```diff
@@ -0,0 +1,5 @@
+#! /bin/bash
```
anorth, line 1: Could you please add a brief description and usage example to these?

anorth (approved):

PR #36, opened 2024-09-23
Proofset ownership transfer

### LINKED ISSUE #25: Ownership management
Since we are managing ownership at the proof set level we need to explicitly implement ownership upgrades.  We can probably use a library to do this. 

We want this so that providers can have the basics of key management.

ZenGround0:
Closes #25

### DIFF
--- a/contracts/src/PDPService.sol
+++ b/contracts/src/PDPService.sol
@@ -55,6 +55,7 @@ contract PDPService {
     // ownership of proof set is initialized upon creation to create message sender 
     // proofset owner has exclusive permission to add and remove roots and delete the proof set
     mapping(uint256 => address) proofSetOwner;
+    mapping(uint256 => address) proofSetProposedOwner;
 
     // Methods
     constructor(uint256 _challengeFinality) {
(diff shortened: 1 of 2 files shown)


contracts/test/PDPService.t.sol
```diff
@@ -70,6 +72,50 @@ contract PDPServiceProofSetCreateDeleteTest is Test {
...
+        pdpService = new PDPService(2);
+
+        owner = address(this);
+        nextOwner = address(0x1234);
+        nonOwner = address(0xffff);
+    }
+
+    function testOwnershipTransfer() public {
+        uint256 setId = pdpService.createProofSet();
+        pdpService.proposeProofSetOwner(setId, nextOwner);
```
anorth, line 91: Check getProofSetOwner here too

contracts/src/PDPService.sol
```diff
@@ -111,6 +112,21 @@ contract PDPService {
         return rootLeafCounts[setId][rootId];
     }
 
+    // owner proposes new owner
+    function proposeProofSetOwner(uint256 setId, address newOwner) public {
+        require(proofSetLive(setId), "Proof set not live");
+        require(proofSetOwner[setId] == msg.sender, "Only the current owner can propose a new owner");
+        proofSetProposedOwner[setId] = newOwner;
```
anorth, line 119: I suggest giving the current owner a way to clear the proposal by proposing self as new owner (which should not require claiming).

anorth (approved):

PR #38, opened 2024-09-23
Record next challenge epoch. Extract CID to library.

anorth:
This is a prefactor extracted from my work implementing `provePosession`. In addition to adding the challenge epoch, I realised there is insufficient testing for the basic logic around adding and removing pieces, and resolving all this was going to balloon the PR adding proving (for which the testing will be quite involved).

I haven't added all the necessary tests here, but it would be great @ZenGround0 if you could follow on from this and fill out those workflow tests.

### DIFF
--- a/contracts/src/Cids.sol
+++ b/contracts/src/Cids.sol
@@ -1,0 +1,40 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.20;
+
+library Cids {
+    // TODO PERF: https://github.com/FILCAT/pdp/issues/16#issuecomment-2329836995 
+    struct Cid {
+        bytes data;
+    }
+
+    function digestFromCid(Cid memory cid) public pure returns (bytes32) {
+        require(cid.data.length >= 32, "Cid data is too short");
+        // bytes memory dataSlice = new bytes(32);
+        // Assembly version of
+        // for (uint i = 0; i < 32; i++) {
+        //     dataSlice[i] = cid.data[cid.data.length - 32 + i];
+        // }
+        // return bytes32(dataSlice);   
+        assembly {
+            // Allocate memory for the data slice.
+            // Set dataSclice to the next 32-byte aligned location after the free memory pointer (at 0x40).
+            let dataSlice := add(mload(0x40), 0x20)
+            // Update free memory pointer to next slot after dataSlice.
+            mstore(0x40, add(dataSlice, 0x20))
+
+            // Copy the last 32 bytes from cid.data to dataSlice.
+            // Load the length of cid.data (length is the first word in an array memory layout).
+            let dataLength := mload(mload(cid))
+            // Calculate starting position of the last 32 bytes in cid.data.
+            let dataStart := add(mload(cid), sub(dataLength, 32))
+            // Copy 32 bytes (one word) from dataStart into dataSlice.
+            mstore(dataSlice, mload(dataStart))
+
+            // Return the 32-byte slice as bytes32.
+            // Store the data slice at 0x0 (an unused memory word).
+            mstore(0x0, mload(dataSlice))
+            // Return the 32 bytes at Ox0 as a bytes32.
+            return(0x0, 0x20)
+        }
+    }
+}
--- a/contracts/src/PDPService.sol
+++ b/contracts/src/PDPService.sol
@@ -2,16 +2,14 @@
 pragma solidity ^0.8.13;
 
 import {BitOps} from "../src/BitOps.sol";
+import {Cids} from "../src/Cids.sol";
+import {MerkleProof} from "../src/Proofs.sol";
 
 contract PDPService {
     // Constants
     uint256 public constant LEAF_SIZE = 32;
 
     // Types
-    // TODO PERF: https://github.com/FILCAT/pdp/issues/16#issuecomment-2329836995 
-    struct Cid {
-        bytes data;
-    }
 
     // State fields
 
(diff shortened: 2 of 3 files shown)


contracts/test/PDPService.t.sol
```diff
@@ -70,6 +78,32 @@ contract PDPServiceProofSetCreateDeleteTest is Test {
...
+        assertEq(pdpService.getProofSetLeafCount(setId), leafCount);
+        assertEq(pdpService.getNextChallengeEpoch(setId), block.number + 2);
+
+        assertTrue(pdpService.rootLive(setId, rootId));
+        assertEq(pdpService.getRootCid(setId, rootId).data, roots[0].root.data);
+        assertEq(pdpService.getRootLeafCount(setId, rootId), leafCount);
+    }
+
+    // TODO: test bad inputs to addRoot
+    // TODO: test removing roots, good and bad inputs
```
ZenGround0, line 104: There already exist tests removing roots though I agree we don't have as much testing as I'd like I thought that this fell more under fuzzing than unit tests. What else are you specifically looking for here?

ZenGround0: (https://github.com/FILCAT/pdp/pull/38/files#diff-5dd05f8e5c9c7b2856abd89d1eaeeafae1f35807caf7369c7f1748df051ce89bR220)

ZenGround0: I think I get it, you'd like a more thorough check of all state variable changed by adding / removing.  Can do

contracts/src/Cids.sol
```diff
@@ -0,0 +1,19 @@
...
+library Cids {
+    // TODO PERF: https://github.com/FILCAT/pdp/issues/16#issuecomment-2329836995 
+    struct Cid {
+        bytes data;
+    }
+
+    function digestFromCid(Cid memory cid) public pure returns (bytes32) {
+        require(cid.data.length >= 32, "Cid data is too short");
+        bytes memory dataSlice = new bytes(32);
+        // This can probably be optimised in assembly.
```
ZenGround0, line 13: Its probably worth doing that here right now.  I remember [playing around with mload](https://github.com/FILCAT/client-contract/blob/zen/dev/src/CBORParse.sol#L95) to do something similar a long time ago and it was pretty simple.

contracts/src/PDPService.sol
```diff
@@ -131,33 +138,39 @@ contract PDPService {
...
         require(rootData.length > 0, "Must add at least one root");
         require(proofSetOwner[setId] == msg.sender, "Only the owner can add roots");
+        bool needsChallengeEpoch = nextChallengeEpoch[setId] == 0;
         uint256 firstAdded = nextRootId[setId];
 
         for (uint256 i = 0; i < rootData.length; i++) {
             addOneRoot(setId, i, rootData[i].root, rootData[i].rawSize);
         }
+        // Initialise the first challenge epoch when the first data is added.
+        if (needsChallengeEpoch && proofSetLeafCount[setId] > 0) {
```
ZenGround0, line 161: I'd prefer dropping the second check.  Since addOneRoot only succeeds if size > 0 it is redundant.

contracts/src/PDPService.sol
```diff
@@ -131,33 +138,39 @@ contract PDPService {
...
 
     error IndexedError(uint256 idx, string msg);
 
     // Appends a new root to the collection managed by a proof set.
     // Must be called by the proof set owner.  
-    function addOneRoot(uint256 setId, uint256 callIdx, Cid calldata root, uint256 rawSize) internal returns (uint256) {
+    function addOneRoot(uint256 setId, uint256 callIdx, Cids.Cid calldata root, uint256 rawSize) internal returns (uint256) {
         if (rawSize % LEAF_SIZE != 0) {
-            revert IndexedError(callIdx, "Size must be a multiple of 256");
+            revert IndexedError(callIdx, "Size must be a multiple of 32");
```
ZenGround0, line 173: Thanks

contracts/src/PDPService.sol
```diff
@@ -184,6 +197,11 @@ contract PDPService {
             totalDelta += removeOneRoot(setId, rootIds[i]);
         }
         proofSetLeafCount[setId] -= totalDelta;
+        // Clear next challenge epoch if the set is now empty.
+        // It will be re-set when new data is added.
+        if (proofSetLeafCount[setId] == 0) {
```
ZenGround0, line 202: I think we should keep a counter of live roots and consult that both for the needsChallengeEpoch bool above and for this one.  I think it makes sense to do here.  But if you'd rather defer that that's ok.

ZenGround0 (approved):

contracts/test/PDPService.t.sol
```diff
@@ -70,6 +78,32 @@ contract PDPServiceProofSetCreateDeleteTest is Test {
...
+
+    function testAddRoot() public {
+        uint256 setId = pdpService.createProofSet();
+        PDPService.RootData[] memory roots = new PDPService.RootData[](1);
+        roots[0] = PDPService.RootData(Cids.Cid(abi.encodePacked("test")), 64);
+        uint256 rootId = pdpService.addRoots(setId, roots);
+        
+        uint256 leafCount = roots[0].rawSize / 32;
+        assertEq(pdpService.getProofSetLeafCount(setId), leafCount);
+        assertEq(pdpService.getNextChallengeEpoch(setId), block.number + 2);
```
ZenGround0, line 96: Probably we should extract this 2 to a testing contract constant for this test

contracts/src/PDPService.sol
anorth: I'm not convinced yet, so will defer.

PR #41, opened 2024-09-24
Implement proof of possesion

### LINKED ISSUE #9: ProvePossession -- challenge index generation
1. Determine the correct randomness to draw from based on challenge period and latest proof of possession
2. Sample randomness from evm GetRandDAO opcode 
3. Use randomness to get challenge index

### LINKED ISSUE #10: ProvePossession -- verify proof
1. Use challenge index with SumTree search (#8 ) to find root to prove against
2. Verify merkle proof argument against root id hash 

Things adding uncertainty
* we should find and import a good merkle proving library which has handled lots of proving.  @nicola has done some investigation into this IIRC.
* we need to figure out cids vs no cids in our proofs

anorth:
Implemented PDPService.provePosession method, with tests.

At present, uses the challenge epoch itself as the random seed rather than fetching chain randomness from that epoch (see #44).

I moved the Merkle tree and proof building code alongside the verification code since now its needed from multiple test files. The compiler should strip un-called functions from these libraries when deploying.

Closes #9. Closes #10.

### DIFF
--- a/contracts/src/Cids.sol
+++ b/contracts/src/Cids.sol
@@ -7,34 +7,49 @@ library Cids {
         bytes data;
     }
 
+    // Returns the last 32 bytes of a CID payload as a bytes32.
     function digestFromCid(Cid memory cid) public pure returns (bytes32) {
         require(cid.data.length >= 32, "Cid data is too short");
-        // bytes memory dataSlice = new bytes(32);
-        // Assembly version of
-        // for (uint i = 0; i < 32; i++) {
-        //     dataSlice[i] = cid.data[cid.data.length - 32 + i];
-        // }
-        // return bytes32(dataSlice);   
-        assembly {
-            // Allocate memory for the data slice.
-            // Set dataSclice to the next 32-byte aligned location after the free memory pointer (at 0x40).
-            let dataSlice := add(mload(0x40), 0x20)
-            // Update free memory pointer to next slot after dataSlice.
-            mstore(0x40, add(dataSlice, 0x20))
-
-            // Copy the last 32 bytes from cid.data to dataSlice.
-            // Load the length of cid.data (length is the first word in an array memory layout).
-            let dataLength := mload(mload(cid))
-            // Calculate starting position of the last 32 bytes in cid.data.
-            let dataStart := add(mload(cid), sub(dataLength, 32))
-            // Copy 32 bytes (one word) from dataStart into dataSlice.
-            mstore(dataSlice, mload(dataStart))
-
-            // Return the 32-byte slice as bytes32.
-            // Store the data slice at 0x0 (an unused memory word).
-            mstore(0x0, mload(dataSlice))
-            // Return the 32 bytes at Ox0 as a bytes32.
-            return(0x0, 0x20)
+        bytes memory dataSlice = new bytes(32);
+        for (uint i = 0; i < 32; i++) {
+            dataSlice[i] = cid.data[cid.data.length - 32 + i];
         }
+        return bytes32(dataSlice);   
+        // The above can almost certainly be optimised in assembly.
+        // However, the below assembly behaves differently and is beyond my ability to resolve.
+        // assembly {
+        //     // Allocate memory for the data slice.
+        //     // Set dataSclice to the next 32-byte aligned location after the free memory pointer (at 0x40).
+        //     let dataSlice := add(mload(0x40), 0x20)
+        //     // Update free memory pointer to next slot after dataSlice.
+        //     mstore(0x40, add(dataSlice, 0x20))
+
+        //     // Copy the last 32 bytes from cid.data to dataSlice.
+        //     // Load the length of cid.data (length is the first word in an array memory layout).
+        //     let dataLength := mload(mload(cid))
+        //     // Calculate starting position of the last 32 bytes in cid.data.
+        //     let dataStart := add(mload(cid), sub(dataLength, 32))
+        //     // Copy 32 bytes (one word) from dataStart into dataSlice.
+        //     mstore(dataSlice, mload(dataStart))
+
+        //     // Return the 32-byte slice as bytes32.
+        //     // Store the data slice at 0x0 (an unused memory word).
+        //     mstore(0x0, mload(dataSlice))
+        //     // Return the 32 bytes at Ox0 as a bytes32.
+        //     return(0x0, 0x20)
+        // }
+    }
+
+    // Makes a CID from a prefix and a digest.
+    // The prefix doesn't matter to these contracts, which only inspect the last 32 bytes (the hash digest).
+    function cidFromDigest(bytes memory prefix, bytes32 digest) internal pure returns (Cids.Cid memory) {
+        bytes memory byteArray = new bytes(prefix.length + 32);
+        for (uint256 i = 0; i < prefix.length; i++) {
+            byteArray[i] = prefix[i];
+        }
+        for (uint256 i = 0; i < 32; i++) {
+            byteArray[i+prefix.length] = bytes1(digest << (i * 8));
+        }
+        return Cids.Cid(byteArray);
     }
 }
--- a/contracts/src/Cids.sol
+++ b/contracts/src/Cids.sol
@@ -15,6 +15,7 @@ library Cids {
             dataSlice[i] = cid.data[cid.data.length - 32 + i];
         }
         return bytes32(dataSlice);   
+        // See https://github.com/FILCAT/pdp/issues/15 performance measurement and improvement
         // The above can almost certainly be optimised in assembly.
         // However, the below assembly behaves differently and is beyond my ability to resolve.
         // assembly {
--- a/contracts/src/PDPService.sol
+++ b/contracts/src/PDPService.sol
@@ -304,6 +304,7 @@ contract PDPService {
     function provePossession(uint256 setId, Proof[] calldata proofs) public {
         uint256 challengeEpoch = nextChallengeEpoch[setId];
         require(block.number >= challengeEpoch, "premature proof");
+        require(proofs.length > 0, "empty proof");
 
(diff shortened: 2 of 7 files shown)


contracts/src/Cids.sol
```diff
@@ -7,34 +7,49 @@ library Cids {
...
-            let dataSlice := add(mload(0x40), 0x20)
-            // Update free memory pointer to next slot after dataSlice.
-            mstore(0x40, add(dataSlice, 0x20))
+        bytes memory dataSlice = new bytes(32);
+        for (uint i = 0; i < 32; i++) {
+            dataSlice[i] = cid.data[cid.data.length - 32 + i];
+        }
+        return bytes32(dataSlice);   
+        // The above can almost certainly be optimised in assembly.
+        // However, the below assembly behaves differently and is beyond my ability to resolve.
```
anorth, line 19: @ZenGround0 I tried assembly after your suggestion in the previous PR, but eventually tracked proof failures down to this. I would rather leave it to an expert. I can delete the incorrect assembly if you wish, though it might provide a head start to someone coming in.

contracts/src/PDPService.sol
```diff
@@ -303,7 +302,29 @@ contract PDPService {
...
+
+        // TODO: fetch proper seed from chain randomness
+        uint256 seed = challengeEpoch;
+        uint256 leafCount = getProofSetLeafCount(setId);
+        uint256 sumTreeTop = 256 - BitOps.clz(nextRootId[setId]);
+
+        for (uint64 i = 0; i < proofs.length; i++) {
+            // Hash (SHA3) the seed,  proof set id, and proof index to create challenge.
+            bytes memory payload = abi.encodePacked(seed, setId, i);
+            uint256 challengeIdx = uint256(keccak256(payload)) % leafCount;
```
anorth, line 316: This is an easy hash to use here, but we could use sha256 if preferred.

contracts/src/Proofs.sol
```diff
@@ -146,6 +147,64 @@ library MerkleProof {
     }
 }
 
+library MerkleProve {
```
anorth, line 150: No changes here, just moved the code.

contracts/src/Cids.sol
ZenGround0: Its way better to keep it commented out, thank you

contracts/src/PDPService.sol
ZenGround0: sha3 lgtm

contracts/src/PDPService.sol
```diff
@@ -303,7 +302,29 @@ contract PDPService {
...
+
+            // Find the root that has this leaf, and the offset of the leaf within that root.
+            RootIdAndOffset memory root = findOneRootId(setId, challengeIdx, sumTreeTop);
```
ZenGround0, line 319: We should be using `findRootIds` here to make use of batching cost reduction.

contracts/test/PDPService.t.sol
```diff
@@ -313,6 +316,133 @@ contract PDPServiceProofSetMutateTest is Test {
...
+    }
+
+    function testProveManyRoots() public {
+        // Build Merkle trees.
+        uint rootCount = 3;
+        uint[] memory leafCounts = new uint[](rootCount);
+        bytes32[][][] memory trees = new bytes32[][][](rootCount);
+        PDPService.RootData[] memory roots = new PDPService.RootData[](rootCount);
+        for (uint i = 0; i < rootCount; i++) {
+            // Generate different size trees up to max leaves.
```
ZenGround0, line 364: The protocol constant for max leaves is 1 << 50 so this comment is confusing

contracts/test/PDPService.t.sol
```diff
@@ -313,6 +316,133 @@ contract PDPServiceProofSetMutateTest is Test {
...
+        for (uint challengeIdx = 0; challengeIdx < challengeCount; challengeIdx++) {
+            // Compute challenge index
+            bytes memory payload = abi.encodePacked(seed, setId, uint64(challengeIdx));
+            uint256 challengeOffset = uint256(keccak256(payload)) % totalLeafCount;
+            // console.log("challengeIdx", vm.toString(challengeIdx));
+
+            uint treeIdx = 0;
+            uint256 treeOffset = 0;
+            for (uint i = 0; i < leafCounts.length; ++i) {
+                if (leafCounts[i] >= challengeOffset) {
```
ZenGround0, line 419: It doesn't look like this is the bug causing failure but I think this <= is off by one and should be <.  If we got unlucky and we landed with a challenge offset == leafCount[i], i.e. a leaf count of 5 and a challenge offset of 5 then we'd try to access leafs[5] which would overrun the leaf buffer.

contracts/test/PDPService.t.sol
```diff
@@ -313,6 +316,133 @@ contract PDPServiceProofSetMutateTest is Test {
...
+                    treeIdx = i;
+                    treeOffset = challengeOffset;
```
ZenGround0, line 421:
```suggestion
                    treeOffset = challengeOffset;
                    break;
```

ZenGround0 (commented):
This should fix it

contracts/src/PDPService.sol
```diff
@@ -303,7 +302,29 @@ contract PDPService {
...
+            bytes32 rootHash = Cids.digestFromCid(getRootCid(setId, root.rootId));
+            bool ok = MerkleVerify.verify(proofs[i].proof, rootHash, proofs[i].leaf, root.offset);
+            require(ok, "proof did not verify");
+        }
+
+        // Set the next challenge epoch.
+        nextChallengeEpoch[setId] = block.number + challengeFinality; 
+        // TODO: record the successful proof somewhere.
```
ZenGround0, line 327: You can remove this as the record keeper PR should reflect these requirements.  I hadn't planned on storing the actual proof but maybe we should at least add it into the recordkeeper call for the consumer to potentially keep around.

contracts/test/ProofUtil.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.20;
+
+import {Cids} from "../src/Cids.sol";
+import {MerkleVerify, Hashes} from "../src/Proofs.sol";
+
+// Methods for committing to data and generating proofs.
+// These are only used in tests (which verify proofs).
+// These functions provide a spec for the operations which providers should perform off-chain.
+library ProofUtil {
```
ZenGround0, line 10: Were you planning to move more stuff here?  If not I don't understand extracting this piece out and leaving the others, I think at least buildProof and buildTree belong here.

ZenGround0 (commented):
I'll hold off on approval until you take it out of draft and add more tests but with bug fixes this is LingGTM

contracts/test/PDPService.t.sol
anorth: I don't want to build a 2^50 merkle tree in a test. I'll clarify this inline constant.

contracts/src/PDPService.sol
anorth: It doesn't look to me like there is any efficiency gain from the batched method. To call it, I'll have to loop twice here and accumulate the roots in memory between those iterations.

contracts/src/PDPService.sol
anorth: I just meant record that the proof happened.

contracts/test/ProofUtil.sol
anorth: The move was strictly based on generateLeaves being shared between multiple test files.

I do see value in moving some other small helpers out here too. `buildProofs` (plural) is tricker to home. It's the spec for how proofs should be built off chain, but specific to the PDP service (rather than generic Merkle tree ops that I moved to Proofs.sol).

contracts/src/Cids.sol
```diff
@@ -7,34 +7,49 @@ library Cids {
...
+    }
 
-            // Return the 32-byte slice as bytes32.
-            // Store the data slice at 0x0 (an unused memory word).
-            mstore(0x0, mload(dataSlice))
-            // Return the 32 bytes at Ox0 as a bytes32.
-            return(0x0, 0x20)
+    // Makes a CID from a prefix and a digest.
+    // The prefix doesn't matter to these contracts, which only inspect the last 32 bytes (the hash digest).
+    function cidFromDigest(bytes memory prefix, bytes32 digest) internal pure returns (Cids.Cid memory) {
```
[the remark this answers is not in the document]
anorth: Added a roundtrip (which fails hard with the bad assembly)

contracts/src/Proofs.sol
```diff
@@ -146,6 +147,64 @@ library MerkleProof {
...
+    // Builds a merkle tree from an array of leaves.
+    // The tree is an array of arrays of bytes32.
+    // The last array is the leaves, and each prior array is the result of the commutative hash of pairs in the previous array.
```
ZenGround0, line 153:
```suggestion
    // The last array is the leaves, and each prior array is the result of the hash of pairs in the previous array.
```

ZenGround0 (approved):

contracts/test/PDPService.t.sol
```diff
@@ -313,6 +316,263 @@ contract PDPServiceProofSetMutateTest is Test {
...
+        vm.roll(nextChallengeEpoch);
+        
+        // The proof for the old challenge epoch should no longer be valid.
+        vm.expectRevert();
+        pdpService.provePossession(setId, proofs);
+    }
+
+    function testBadRootsRejected() public {
+        uint[] memory leafCounts = new uint[](2);
+        // Note: either co-prime leaf counts or a challenge count > 1 are required for this test to demonstrate the failing proof.
```
ZenGround0, line 417: I'm not sure about this but my best understanding is the co prime condition needs to exist over leaf count1 and (leaf count1 + leaf count2)

PR #43, opened 2024-09-25
Integrate record keeper

### LINKED ISSUE #12: Write to record keeping structure during methods
Prerequisites of this work are 
1. #11 The record keeping data structure
2. At least stub functions existing for all proof set operations.  Probably should block on #7 #5 #6 #10. 

Plumb events describing the proof set operations into writes to the record keeping data structure.  This should be done in a function / abstracted way to prepare for the next phase of record keeping.

### LINKED ISSUE #31: Emit Events when the proof set is updated
Emit events when the proof set is updated, this could make those who are building tools and oracles around data onboard to filecoin easier, like
- register deal/piece for retrieval checks, i.e spark
- CID checker/dashboard tools, like filecoin.tools 
- SP software to recover indexing

ZenGround0:
Closes #12, Closes #31
TODO
- [x] Add assertions of record keeping state to the end of all PDPService tests (create helpers for this) 
- [x] Add tests with faked PDPService contract that fails in different scenarios to ensure failures propagate
- [x] Update deploy shell script to deploy a record keeper 
- [x] Update createproofset shell script to include address of record keeper

### DIFF
--- a/contracts/src/PDPService.sol
+++ b/contracts/src/PDPService.sol
@@ -4,6 +4,7 @@ pragma solidity ^0.8.13;
 import {BitOps} from "../src/BitOps.sol";
 import {Cids} from "../src/Cids.sol";
 import {MerkleVerify} from "../src/Proofs.sol";
+import {PDPRecordKeeper} from "../src/PDPRecordKeeper.sol";
 
 contract PDPService {
     // Constants
@@ -28,7 +29,8 @@ contract PDPService {
         uint256 leafCount;
         address owner;
         nextRootID uint64;
-        nextChallengeEpoch: uint64
+        nextChallengeEpoch: uint64;
+        recordKeeper: address;
     }
     ** PDP service contract tracks many possible proof sets **
     []ProofSet proofsets
@@ -53,10 +55,12 @@ contract PDPService {
     mapping(uint256 => uint256) nextRootId;
     mapping(uint256 => uint256) proofSetLeafCount;
     mapping(uint256 => uint256) nextChallengeEpoch;
+    mapping(uint256 => address) proofSetRecordKeeper;
     // ownership of proof set is initialized upon creation to create message sender 
     // proofset owner has exclusive permission to add and remove roots and delete the proof set
     mapping(uint256 => address) proofSetOwner;
     mapping(uint256 => address) proofSetProposedOwner;
+    
 
     // Methods
     constructor(uint256 _challengeFinality) {
@@ -143,11 +147,14 @@ contract PDPService {
     // A proof set is created empty, with no roots. Creation yields a proof set ID 
     // for referring to the proof set later.
     // Sender of create message is proof set owner.
-    function createProofSet() public returns (uint256) {
+    function createProofSet(address recordKeeper) public returns (uint256) {
         uint256 setId = nextProofSetId++;
         proofSetLeafCount[setId] = 0;
         nextChallengeEpoch[setId] = 0;  // Re-initialized when the first root is added.
         proofSetOwner[setId] = msg.sender;
+        proofSetRecordKeeper[setId] = recordKeeper;
+
+        _addRecord(setId, recordKeeper, PDPRecordKeeper.OperationType.CREATE, bytes(""));
         return setId;
     }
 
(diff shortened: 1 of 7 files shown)


contracts/test/PDPService.t.sol
```diff
@@ -7,14 +7,16 @@ import {Cids} from "../src/Cids.sol";
 
 contract PDPServiceProofSetCreateDeleteTest is Test {
     PDPService pdpService;
+    PDPRecordKeeper recordKeeper;
 
     function setUp() public {
         pdpService = new PDPService(2);
+        recordKeeper = new PDPRecordKeeper(address(pdpService));
```
ZenGround0, line 14: As I'll be adding assertions on the state of the recordkeeper at the end of each test to finish up this PR it is somewhat the case that PDPService tests are becoming too coupled to the PDPRecordKeeper contract.  However my thinking is that 
* The current PDPRecordKeeper which keeps everything matches perfectly with how I would implement fake record keeper
* If we end up tweaking PDPRecordKeeper to be smarter and more efficient I'll be keeping the existing one around for testing purposes so no future improvements to PDPRecordKeeper will impact tests

contracts/src/PDPService.sol
```diff
@@ -184,6 +193,9 @@ contract PDPService {
         if (needsChallengeEpoch) {
             nextChallengeEpoch[setId] = block.number + challengeFinality; 
         }
+
+        bytes memory extraData = abi.encode(rootData.length, firstAdded, rootData);
```
ZenGround0, line 197: I liked putting length first as it makes things easy to parse but realistically its probably overkill since any parsing of rootData will have the length.  What do you think @anorth ?

contracts/src/PDPService.sol
```diff
@@ -368,4 +387,9 @@ contract PDPService {
         return BitOps.ctz(index + 1);
     }
 
+    /* Record keeper functions */
+    function _addRecord(uint256 proofSetId, address recordKeeper, PDPRecordKeeper.OperationType operationType, bytes memory extraData) internal {
```
anorth, line 391: Check for zero record keeper address and don't call it if so

contracts/src/PDPService.sol
```diff
@@ -143,11 +147,14 @@ contract PDPService {
...
     // Sender of create message is proof set owner.
-    function createProofSet() public returns (uint256) {
+    function createProofSet(address recordKeeper) public returns (uint256) {
         uint256 setId = nextProofSetId++;
         proofSetLeafCount[setId] = 0;
         nextChallengeEpoch[setId] = 0;  // Re-initialized when the first root is added.
         proofSetOwner[setId] = msg.sender;
+        proofSetRecordKeeper[setId] = recordKeeper;
+
+        _addRecord(setId, recordKeeper, PDPRecordKeeper.OperationType.CREATE, bytes(""));
```
anorth, line 157: I *think* we'll want the owner address in extra data.

contracts/src/PDPService.sol
```diff
@@ -162,6 +169,8 @@ contract PDPService {
         proofSetLeafCount[setId] = 0;
         proofSetOwner[setId] = address(0);
         nextChallengeEpoch[setId] = 0;
+
+        _addRecord(setId, proofSetRecordKeeper[setId], PDPRecordKeeper.OperationType.DELETE, bytes(""));
```
anorth, line 173: Payload of size/leafcount?

contracts/src/PDPService.sol
anorth: We probably don't need the length.

contracts/src/PDPService.sol
```diff
@@ -325,6 +343,8 @@ contract PDPService {
 
         // Set the next challenge epoch.
         nextChallengeEpoch[setId] = block.number + challengeFinality; 
+        bytes memory extraData = abi.encode(seed);
```
anorth, line 346: From our discussions about single owner versus mixed proof sets, it's important for the event observer to know the total size of the proof set when it's proven. Are you relying on them tracking this from the sequence of adds and removes, including of data that is not of interest? Perhaps the convenience of including the total size with the proof event is worth the overhead.

anorth (approved):

contracts/src/PDPService.sol
ZenGround0: > Are you relying on them tracking this from the sequence of adds and removes, including of data that is not of interest?

Yes I think this is essentially required either after the fact or with a PDP application using logic to guard what goes into the proof set.  But I think this is fine metadata to include here too.

PR #46, opened 2024-10-02
Fix grinding in PDP Service

ZenGround0:
This PR fixes the lack of grinding protection in PDPService.  This PR started out as a spike exploring a couple possibility.  See below for the historical context

--history--
This is a spike to see how we like changing the PDPService directly to disable grinding.

The PDP Service itself only issues challenges against the proof set data that existed before the challenge was generated.  All removals are enqueued and processed after proof so that they cannot be used to selectively grind.  All adds are enqueued and committed to the proving set after proof.

This is the obvious way to do it.  @nicola suggested putting this outside the main service since in cases where the application specifies addition and removal very tightly (i.e. up to ordering and epoch of add and remove) there is no grinding concern.


contracts/src/PDPService.sol
```diff
@@ -314,6 +325,25 @@ contract PDPService {
         bytes32[] proof;
     }
 
+    function proveAndUpdate(uint256 setId, Proof[] calldata proofs) public {
```
ZenGround0, line 328: Other names:
* Prove
* FinishProvingPeriod
* ProofTick
* ProveAndRefresh

Right now I'm leaning towards just `Prove`

contracts/src/PDPService.sol
```diff
@@ -225,9 +230,15 @@ contract PDPService {
         return rootId;
     }
 
+    function enqueueRemovals(uint256 setId, uint256[] calldata rootIds) public {
```
ZenGround0, line 233: Other names

* ProposeRemove
* RegisterRemove
* Remove

Probably we don't want just `Remove` as it gives the wrong idea

contracts/src/PDPService.sol
```diff
@@ -225,9 +230,15 @@ contract PDPService {
...
+        for (uint256 i = 0; i < rootIds.length; i++){
+            enqueuedRemovals[setId].push(rootIds[i]);
```
ZenGround0, line 235: TODO: add application notify call.  This is the only place for access control on removal since once its already slotted for removal proving won't be possible since prove and remove happen atomically.

ZenGround0: TODO: probably want limits on how many removals we can enqueue so that proof of possession doesn't run out of gas

contracts/src/PDPService.sol
```diff
@@ -191,9 +195,10 @@ contract PDPService {
         for (uint256 i = 0; i < rootData.length; i++) {
             addOneRoot(setId, i, rootData[i].root, rootData[i].rawSize);
```
ZenGround0, line 196: TODO: write comment on add function that this data is not going to be proven until the next challenge sampling event

contracts/src/PDPService.sol
ZenGround0, line 233: TODO: probably check here that proofset is not empty

ZenGround0:
I've decided this is the right way forward so I'll convert this spike into a real PR shortly

anorth:
A big issue I see with this approach is that if a provider is somehow unable to prove a challenged leaf, for example, they've lost the data, then the entire proof set is permanently borked. 

We need to allow the provider to admit defeat and move on somehow. The prior code does this by allowing the SP to add or remove roots to create a challenge they can meet, i.e. grinding, but in such a way that might fail the application level requirements. I agree we do need to address grinding more directly though and wish that had been considered in the protocol design earlier.

ZenGround0:
> We need to allow the provider to admit defeat and move on somehow. 

Great point I hadn't uncovered yet.  Right now I'm thinking "faultProvingPeriod" which allows processing all proving period operations including removals and resampling and registering a fault.

ZenGround0:
@anorth your comments got me experimenting with a better factoring which I'm convinced is better than before.  We now have two methods on the interface in place of the one completeProvingPeriod.
1. The old `provePossess` except it does not sample challenges
2. A new `nextProvingPeriod` which updates the proving set and resamples challenges.

For liveness nextProvingPeriod can be called without proving first.  The application sitting on the receiver can make note of this and fault.

It might be cleaner to force an original call to nextProvingPeriod before the first provePossession instead of having the first add setup the first sampling. 

I can see the case that its a nice optimization to call nextProvingPeriod at the end of provePossession.  But I think its better for everyone to keep them decoupled.

contracts/src/PDPService.sol
```diff
@@ -308,12 +347,45 @@ contract PDPService {
...
+    // 1. Resample challenge
+    // 2. Track newly added roots for proving
+    // 3. Remove scheduled removals
+    function nextProvingPeriod(uint256 setId) public {
+        require(msg.sender == proofSetOwner[setId], "only the owner can move to next proving period");
+        // Take removed roots out of proving set
+        uint256[] storage removals = scheduledRemovals[setId];
+        uint256[] memory removalsToProcess = new uint256[](removals.length);
+    
+        for (uint256 i = 0; i < removalsToProcess.length; i++) {
```
magik6k, line 365: Should there be some limit on how many removals can be processed at once? If there isn't a limit it looks like it is possible to brick a proving set with too many removals

anorth (commented):
I'm expecting you rebase and to push a few more changes in here after we talked in person. So this wasn't yet a complete in-depth review.

contracts/src/PDPRecordKeeperApplication.sol
```diff
@@ -1,36 +1,47 @@
...
+import {PDPService} from "./PDPService.sol";
 
-    // Enum to represent different types of operations
+interface PDPApplication {
     enum OperationType {
         NONE,
         CREATE,
         ADD,
-        REMOVE,
+        REMOVE_SCHEDULED,
```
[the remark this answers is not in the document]
ZenGround0: I think we need to hear about removal schedule otherwise the listener can't fail on removal without breaking liveness

contracts/src/PDPService.sol
ZenGround0: Yeah out of concern for this I put this check here: https://github.com/FILCAT/pdp/pull/46/files#diff-712b2db21b090e745391534efb9ce9dfb70c3f729952e9433b627e264e3f83a3R261.  One of the things I'm most worried about with this change is how to set this number.  I arbitrarily set it to 2000 after eyeballing batch removal gas costs.  But its possible this could be either annoyingly low or dangerously high.

contracts/src/PDPRecordKeeperApplication.sol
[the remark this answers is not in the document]
ZenGround0: For example imagining an application like f05 as the listener you'd want to make sure that you are allowed to remove a cid from the proofset because the deal is expired.

contracts/src/PDPService.sol
```diff
@@ -181,6 +212,8 @@ contract PDPService {
         uint256 rawSize;
     }
 
+    // Appends new roots to the collection managed by a proof set.
+    // These roots won't be selected for proving until the next proving period.
```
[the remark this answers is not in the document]
ZenGround0: Lets think about this thoroughly.  There is state asymmetry
### State that `add` updates eagerly and `scheduleRemove` updates lazily
 - sumTree
 - leafCounts
 - leafCount
 - nextRootID

### State both update lazily
- lastChallengedLeaf

### Abstract asymmetry
Both new add and scheduleRemove are invisible to provePossession in terms of correctness and visible by performance though add is much more perf visible via sumtree.  That's probably the biggest abstract asymmetry I see but it only shows up through gas cost.

New adds are visible to other adds.  We need this for eager sequencing which is necessary.   New scheduleRemovals are not visible to other adds but adds never depended on any previous removes because proofsets are append only.  So there's no new symmetry break.

New adds are visible to scheduleRemoves in that when scheduled removals are processed they can remove new adds that have not yet been put into the proving set.  This is a better way to do things imo than to require all new adds are put in the proof set for one proving period before they can be deleted. Removes don't need to consider other removes because remove operation is idempotent.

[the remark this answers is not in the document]
ZenGround0: State asymmetry in sumtree, leafcounts and leafCount could be replaced by enqueuing add operations for processing in nextProvingPeriod like removals.  This would also get rid of the big gas cost asymmetry since sumtree find cost only changes each proving period. 

nextRootID state asymmetry exists to support good add sequencing.  If we enqueued adds we could instead update this only in nextProvingPeriod and check the length of the queue to return correct sequence numbers.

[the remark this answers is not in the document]
ZenGround0: I think the trade off is asymmetry in order to do adds eagerly.  I prefer eager operations for user flexibility unless its impossible to avoid them, as in the case of removals.  So I think the asymmetry is worth it.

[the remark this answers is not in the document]
ZenGround0: Well one pareto improvement is maybe to keep a variable for uncommitted leaf count and then add this to leafCounts at the end of nextProvingPeriod.  

Then we get rid of "lastChallengedLeaf" and prove against leafCount.

ZenGround0:
>  push a few more changes in here after we talked in person.

FYI going off of @magik6k 's suggestion we'll keep initial sampling on addRoots.  I think its a small bit worse but I'm practicing listening to users and if it gets too annoying later it will be simple to change.

contracts/src/PDPService.sol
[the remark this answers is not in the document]
ZenGround0: No I actually don't like this because the sumtree needs a consistent view of the total leaves in order to work well so we'd need to add the uncommitted leaf count to total leafCount each time we ran a sumtree operation which seems worse.

contracts/src/PDPRecordKeeperApplication.sol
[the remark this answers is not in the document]
anorth: Oooh, we're running into similar issues as the sector activation event. I hadn't thought about the listener failing and being able to prevent the SP from performing some operations. My first thought is that the PDP service should catch and ignore listener failures. SLA enforcement in this way wasn't the goal. That's a good discussion for us this week.

contracts/src/PDPService.sol
```diff
@@ -80,6 +85,8 @@ contract PDPService {
     mapping(uint256 => uint256) nextChallengeEpoch;
     // Each proof set notifies a configurable listener to implement extensible applications managing data storage
     mapping(uint256 => address) proofSetListener;
+    mapping(uint256 => uint256) lastChallengedLeaf;
```
anorth, line 88: Nit: I think we can improve this name. On the surface it sounds like the index of the leaf that was most recently actually challenged (i.e. after sampling randomness and performing modulus). The "last" and past tense here conflicts with "next" challenge epoch, where they should be the same thing (I know one is time vs one is space). Suggestions.
- challengeRange
- challengeLeafCount

In all cases it probably deserves a comment.

contracts/src/PDPService.sol
```diff
@@ -147,6 +154,29 @@ contract PDPService {
         return rootLeafCounts[setId][rootId];
     }
 
+    // Returns the last challenged leaf for a given proof set
```
anorth, line 157: See comment about renaming, but I think the comment should be "challenge-able"?

contracts/src/PDPService.sol
```diff
@@ -147,6 +154,29 @@ contract PDPService {
...
+    function getLastChallengedLeaf(uint256 setId) public view returns (uint256) {
+        require(proofSetLive(setId), "Proof set not live");
+        return lastChallengedLeaf[setId];
+    }
+
+    // Returns the scheduled removals for a given proof set
```
anorth, line 163: Returns the indices of roots that are scheduled for removal for the next proving period.

contracts/src/PDPService.sol
```diff
@@ -207,6 +237,8 @@ contract PDPService {
         uint256 rawSize;
     }
 
+    // Appends new roots to the collection managed by a proof set.
+    // These roots won't be selected for proving until the next proving period.
```
anorth, line 241:
```suggestion
    // These roots won't be challenged until the next proving period.
```

Just for consistent terminology?

contracts/src/PDPService.sol
```diff
@@ -253,31 +284,36 @@ contract PDPService {
...
+        for (uint256 i = 0; i < rootIds.length; i++){
+            totalDelta += rootLeafCounts[setId][rootIds[i]];
+            scheduledRemovals[setId].push(rootIds[i]);
+        }
+        
+        bytes memory extraData = abi.encode(totalDelta, rootIds);
+        _notifyListener(setId, proofSetListener[setId], PDPListener.OperationType.REMOVE_SCHEDULED, extraData);
+    }
+
     // removeRoots removes a batch of roots from a proof set.  Must be called by the proof set owner.
```
anorth, line 301: Move and adapt this comment up to the public scheduleRemovals.

I would also suggest moving the internal remove roots functions down below nextProvingPeriod.

contracts/src/PDPService.sol
```diff
@@ -253,31 +284,36 @@ contract PDPService {
...
-        bytes memory extraData = abi.encode(totalDelta, rootIds);
-        _notifyListener(setId, proofSetListener[setId], PDPListener.OperationType.REMOVE, extraData);
-        emit RootsRemoved(totalDelta);
 
         return totalDelta;
     }
 
     // removeOneRoot removes a root from a proof set. 
     function removeOneRoot(uint256 setId, uint256 rootId) internal returns (uint256) {
+        emit Debug("removing root", rootId);
```
anorth, line 316: Remove the debug before merging?

contracts/src/PDPService.sol
```diff
@@ -343,6 +379,40 @@ contract PDPService {
         bytes32[] proof;
     }
 
+    // Roll over to the next proving period
+    // 1. Resample challenge
+    // 2. Track newly added roots for proving
+    // 3. Remove scheduled removals
```
anorth, line 385: Expand this documentation to address the caller (SP). Explain that it processes removals and expands the challenge range. Explain also that it can be called without submitting a proof beforehand. But that listener contracts may consider this as some kind of failure.

contracts/src/PDPService.sol
```diff
@@ -253,31 +284,36 @@ contract PDPService {
...
     }
 
+    function scheduleRemovals(uint256 setId, uint256[] calldata rootIds) public {
+        require(rootIds.length + scheduledRemovals[setId].length <= MAX_ENQUEUED_REMOVALS, "Too many removals wait for next proving period to schedule");
+        require(proofSetOwner[setId] == msg.sender, "Only the owner can schedule removal of roots");
+        require(proofSetLive(setId), "Proof set not live");
+        uint256 totalDelta = 0;
+        for (uint256 i = 0; i < rootIds.length; i++){
+            totalDelta += rootLeafCounts[setId][rootIds[i]];
+            scheduledRemovals[setId].push(rootIds[i]);
```
anorth, line 294: A removal of the _next_ root id followed by an add that takes that root id would result in the (delayed) removal trashing the subsequent add. And as bonus returning the wrong delta here (it can't know).

I think this needs to check that the scheduled-to-remove IDs are <= the next root ID at least. There is still scope for this return value to be misleading. For example, repeatedly removing the same root. Perhaps duplicates should also be caught here (by iterating the array?).

Please add tests for these cases.

Sending the notification only when actually removing the root could simplify things (but affects the current ability of the listener to abort and prevent the removal)

contracts/test/PDPService.t.sol
```diff
@@ -348,21 +376,65 @@ contract PDPServiceProofSetMutateTest is Test {
         recordAssert.expectRecord(PDPListener.OperationType.ADD, setId);
         uint256[] memory toRemove = new uint256[](1);
         toRemove[0] = 1;
-        pdpService.removeRoots(setId, toRemove);
-
         toRemove[0] = 0;
```
anorth, line 379: One of these two lines is now redundant, and possibly making the test pointless. It's not obvious from the test itself which one (the reader needs to know what the first root ID allocated is).

Should this just do the original pattern of two (schedule) removals?

contracts/test/PDPService.t.sol
```diff
@@ -348,21 +376,65 @@ contract PDPServiceProofSetMutateTest is Test {
...
+        recordAssert.expectRecord(PDPListener.OperationType.REMOVE_SCHEDULED, setId);
+        // Actual removal does not fail
+        pdpService.nextProvingPeriod(setId); 
+        recordAssert.expectRecord(PDPListener.OperationType.NEXT_PROVING_PERIOD, setId);
+
         assertEq(false, pdpService.rootLive(setId, 0));
 
         // Removing again fails
-        pdpService.removeRoots(setId, toRemove);
+        pdpService.scheduleRemovals(setId, toRemove);
```
anorth, line 389: This comment doesn't make sense of the code (already before this PR)

contracts/test/PDPService.t.sol
```diff
@@ -289,17 +311,21 @@ contract PDPServiceProofSetMutateTest is Test {
...
         recordAssert.expectRecord(PDPListener.OperationType.ADD, setId);
+        assertEq(pdpService.getNextChallengeEpoch(setId), block.number + challengeFinalityDelay);
 
         // Remove root
         uint256[] memory toRemove = new uint256[](1);
         toRemove[0] = 0;
-        pdpService.removeRoots(setId, toRemove);
-        recordAssert.expectRecord(PDPListener.OperationType.REMOVE, setId);
+        pdpService.scheduleRemovals(setId, toRemove);
+        recordAssert.expectRecord(PDPListener.OperationType.REMOVE_SCHEDULED, setId);
```
anorth, line 320: I think we need to explicitly check the extra data associated with the event. It's a critical part of correctness for applications built on this.

contracts/src/PDPService.sol
```diff
@@ -253,31 +284,36 @@ contract PDPService {
...
+        }
+        
+        bytes memory extraData = abi.encode(totalDelta, rootIds);
+        _notifyListener(setId, proofSetListener[setId], PDPListener.OperationType.REMOVE_SCHEDULED, extraData);
```
anorth, line 298: Return the total delta too?

contracts/test/PDPService.t.sol
anorth: Consider asserting that the root is not yet removed in any of the methods where this could be observed.

contracts/test/PDPService.t.sol
```diff
@@ -488,41 +568,51 @@ contract PDPServiceProofTest is Test {
...
         vm.roll(challengeEpoch);
         uint challengeCount = 1;
         PDPService.Proof[] memory proofsOneRoot = buildProofsForSingleton(setId, challengeCount, trees[0], leafCounts[0]);
 
-        // Add another root before submitting the proof.
-        uint256 newRootId;
-        (trees[1], newRootId) = addOneRoot(setId, leafCounts[1]);
-        
         // The proof for one root should be invalid against the set with two.
+        console.log("The proof for one root should be invalid against the set with two.");
```
anorth, line 582: Remove logs before merging

contracts/src/PDPService.sol
ZenGround0: Thanks
Off topic but I just discovered this way of debugging solidity contracts and it works really nicely

contracts/src/PDPService.sol
ZenGround0: Great points

I was back and forth between checking if root has been added before deleting.  I think it is better to do that check.

The more I think about it the total delta was just a guess about what the user wanted.  But the listener has all this information from addRoots. I'm more inclined to allow duplicates in the queue and drop any information about the delta.  This then makes removal match add better.

contracts/test/PDPService.t.sol
ZenGround0: I think you're saying this is important for all of our expectRecord checks.  If so I agree.  But I'll wait for #55 and build a framework `expectScheduleRemove` `expectProvePossesion` etc.

contracts/src/PDPService.sol
anorth: Dropping delta SGTM

contracts/src/PDPService.sol
anorth: Ignore this when you stop sending it to the listener.

contracts/test/PDPService.t.sol
ZenGround0: And yeah I think this was just a mistake

contracts/test/PDPService.t.sol
```diff
@@ -338,31 +366,96 @@ contract PDPServiceProofSetMutateTest is Test {
...
-        recordAssert.expectRecord(PDPListener.OperationType.REMOVE, setId);
-        assertEq(false, pdpService.rootLive(setId, 0));
+        // Scheduling an un-added root for removal should fail
+        toRemove[0] = 0; // current (unchallengeable) root
+        toRemove[1] = 1; // future root
+        vm.expectRevert("Can only schedule removal of existing roots");
+        pdpService.scheduleRemovals(setId, toRemove);
+        recordAssert.expectRecord(PDPListener.OperationType.REMOVE_SCHEDULED, setId);
+        // Actual removal does not fail
+        pdpService.nextProvingPeriod(setId); // root 0 is now challengable
```
anorth, line 387: But was it removable before it was challengable? This test doesn't exercise that.

contracts/test/PDPService.t.sol
```diff
@@ -338,31 +366,96 @@ contract PDPServiceProofSetMutateTest is Test {
...
         uint256 setId = pdpService.createProofSet(address(recordKeeper));
         recordAssert.expectRecord(PDPListener.OperationType.CREATE, setId);
         PDPService.RootData[] memory roots = new PDPService.RootData[](1);
         roots[0] = PDPService.RootData(Cids.Cid(abi.encodePacked("test")), 64);
         pdpService.addRoots(setId, roots);
         recordAssert.expectRecord(PDPListener.OperationType.ADD, setId);
-        uint256[] memory toRemove = new uint256[](1);
-        toRemove[0] = 1;
-        pdpService.removeRoots(setId, toRemove);
+        assertEq(true, pdpService.rootChallengable(setId, 0));
```
anorth, line 377: Assert root 1 is not challengable

anorth (approved):

contracts/test/PDPService.t.sol
ZenGround0: I think it does if you see the case of root id 1 below

contracts/test/PDPService.t.sol
ZenGround0: Good point, but it makes more sense to assert its not live as that's the condition that causes failure on removal schedule.

PR #47, opened 2024-10-02
SPIKE: introduce SLA wrapper and put grinding checks there 

ZenGround0:
This delegates grinding fix to another contract we're calling the "SLA"

It keeps the bare PDPService contract very simple at the cost of adding another level of complexity.

### DIFF
--- a/contracts/src/PDPService.sol
+++ b/contracts/src/PDPService.sol
@@ -5,6 +5,59 @@ import {BitOps} from "../src/BitOps.sol";
 import {Cids} from "../src/Cids.sol";
 import {MerkleVerify} from "../src/Proofs.sol";
 import {PDPRecordKeeper} from "../src/PDPRecordKeeper.sol";
+import {Ownable} from "../lib/openzeppelin-contracts/contracts/access/Ownable.sol";
+
+// We could make this a singleton that registers proof sets  
+// but for this spike just keep it very simple -- one SLA contrat per proof set
+contract PDPSLA is Ownable {
+    uint256 lastChallengedLeaf;
+    uint256 untrackedLeafCount;
+    uint256[] enqueuedRemovals;
+    uint256 setId;
+    PDPService pdpService;
+
+    // Owner is allowed to add, remove and prove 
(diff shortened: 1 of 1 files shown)


ZenGround0:
@anorth I'm very curious how this lines up with your idea for PDP applications, whether you'd merge sla and recordkeeper, etc

ZenGround0:
A few things this has helped me sort out. First "SLA" here should be called application.  Second there are two different approaches to SLA contracts -- wrappers using PDP Service as callee and framework where PDP Service is caller.  

We need to choose one approach and stick with it rather than having two as in this experiment.  We've already chosen framework since the record keeper is really just a simple application of the PDP service but we can back out of this if it ends up looking worse.

As to whether we should fix grinding in the application as in this PR vs in PDP Service, we should do it in PDP Service.  All provider initiated removals will need to allow for some flexibility in when an expired root is removed from a proofset.  If that flexibility is allowed to change proof set contents after a challenge is sampled it will degrade security.  

Client initiated removals would be another way to avoid grinding but it is not the intention for how PDP Service is used.  PDP Service is called by provider to perform actions that satisfy the application.  However even if we did go back and change our intent then any reasonable client initiated removals would need to have coordination with the provider.  It would be bad design if client were to call remove the epoch before a proof was due and change the necessary merkle proof inputs.  Client initiated removals would end up needing a coordination mechanism with similar properties to the deferred adds and removals of the grinding solution.

ZenGround0:
Paraphrasing some useful feedback from @magik6k about application as callee vs caller from POV of mining software.

1. The SP doesn’t need to trust the application as much with application as callee. There's a more limited set of problems you need to audit against.  You know for sure that the PDPService contract is getting called when you submit a message.
2. Its easier to have just one contract to call.

anorth (commented):
This has the same issue as #46 if the provider is permanently unable to complete a particular proof.

In general I think it's simpler if we only do PDP -> Application calls for our MVP demonstration, but I would expect more sophisticated things to wrap like this in the future (and call createProofSet and act as the owner).

ZenGround0:
We're abandoning this direction but plan to keep the branch around to document the experiment.

PR #49, opened 2024-10-04
Rename record keeper as application

ZenGround0:
This is a light refactor as the outcome of exploring what PDP application conventions could look like.  This change cements the decision for applications to interact with PDP by getting called back with different operations.

Alternatively we could have had an interface with separate methods for notifying each operation but I didn't see a compelling enough case for this.

### DIFF
--- a/contracts/src/PDPRecordKeeperApplication.sol
+++ b/contracts/src/PDPRecordKeeperApplication.sol
@@ -1,15 +1,9 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.20;
 
-// PDPRecordKeeper is a default implementation of a record keeper for the PDP service.
-// It maintains a record of all events that have occurred in the PDP service,
-// and provides a way to query these events.
-// This contract only supports one PDP service caller, set in the constructor.
-contract PDPRecordKeeper {
-    // The address of the PDP service contract that is allowed to call this contract
-    address public immutable pdpServiceAddress;
+import {PDPService} from "./PDPService.sol";
 
-    // Enum to represent different types of operations
+interface PDPApplication {
     enum OperationType {
         NONE,
         CREATE,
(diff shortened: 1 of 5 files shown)


contracts/src/PDPRecordKeeperApplication.sol
```diff
@@ -19,18 +13,34 @@ contract PDPRecordKeeper {
         DELETE
     }
 
+    function notify(
```
anorth, line 16: `receive`, `receiveProofSetEvent`, `proofSetChanged`, ... ?

contracts/src/PDPService.sol
```diff
@@ -4,15 +4,15 @@ pragma solidity ^0.8.13;
 import {BitOps} from "../src/BitOps.sol";
 import {Cids} from "../src/Cids.sol";
 import {MerkleVerify} from "../src/Proofs.sol";
-import {PDPRecordKeeper} from "../src/PDPRecordKeeper.sol";
+import {PDPApplication} from "../src/PDPRecordKeeperApplication.sol";
```
anorth, line 7: I would try to avoid this circular dependency. I think you can do so by moving the operation type and so on into this contract. It makes sense to me that applications depend on details about how this contract works but not the other way around.

contracts/src/PDPService.sol
```diff
@@ -55,7 +55,8 @@ contract PDPService {
     mapping(uint256 => uint256) nextRootId;
     mapping(uint256 => uint256) proofSetLeafCount;
     mapping(uint256 => uint256) nextChallengeEpoch;
-    mapping(uint256 => address) proofSetRecordKeeper;
+    // Each proof set notifies a configurable application managing data storage
+    mapping(uint256 => address) proofSetApplication;
```
anorth, line 59: I would just call this a receiver or a listener. This contract doesn't care what it is or how it works. This naming approach then applies to many changes below, too.

contracts/test/PDPRecordKeeper.t.sol
```diff
@@ -2,15 +2,15 @@
 pragma solidity ^0.8.13;
```
anorth, line 2: Rename the file to match the contract its testing

anorth (commented):
👍 for the thrust of this. I would name things a little differently, but fundamentally the same outcome. Take what you wish from my suggestions.

anorth (approved):

contracts/src/PDPRecordKeeperApplication.sol
ZenGround0: receiveProofSetEvent

contracts/src/PDPService.sol
ZenGround0: listener is way better

contracts/test/PDPRecordKeeper.t.sol
ZenGround0: Along these lines I changed contract name back to PDPRecordKeeper

PR #51, opened 2024-10-08
Burning fee function

nijoe1:
- Added the required contracts from filecoin-solidity to be able to burn fees on FVM from FEVM
- Added the burn fee function in the PDPService contract
- Created a script to test that burnFee works


contracts/src/PDPService.sol
```diff
@@ -5,11 +5,31 @@ import {BitOps} from "../src/BitOps.sol";
...
+import {SendAPI} from "./lib/SendAPI.sol";
+import {CommonTypes} from "./lib/types/CommonTypes.sol";
+
 
 contract PDPService {
     // Constants
     uint256 public constant LEAF_SIZE = 32;
-    uint256 public constant MAX_ROOT_SIZE = 1 << 50;
+    uint256 public constant MAX_ROOT_SIZE = 1 << 50; 
+    uint64 public immutable BURNER = 0;
```
ZenGround0, line 16: This should be 99 I think

ZenGround0 (commented):
Please just include the last commit so we can evaluate changes separately

ZenGround0:
@nijoe1 I'm realizing that we can simplify this down a lot.  We don't need to import any libraries.  All we need is to define a constant 
```solidity
    address constant BURN_ADDRESS = 0xFf00000000000000000000000000000000000099;
```
and send to that in the burn function.  A regular solidity send should just work without resorting to send functions in filecoin-solidity.  LMK if you have any questions

ZenGround0:
The above has a silly mistake: 0x99 == 153 in decimal

If we define it like this everything should work
```solidity
    address constant BURN_ADDRESS = 0xFf00000000000000000000000000000000000063;
```

ZenGround0 (approved):
🎉

Issue #44, opened 2024-09-26
Fetch proper randomness seed from chain


anorth:
#41 implements proof verification using the challenge epoch itself as the random seed. Fetch randomness instead from the new syscall introduced in builtin-actors, when it's available.

PR #54, opened 2024-10-10
Exercise Precompile 

ZenGround0:
Now that there is a lotus RC with the nv24 candidate we can check in to confirm that it's working.  This contract just returns the precompile output for provided inputs.

This is the first step towards #44


ZenGround0:
We can use this code as part of #44 when that becomes possible to deploy.  For now I'll close this down.

Issue #23, opened 2024-08-28
Upgrade path


ZenGround0:
Idea on upgrade path from @Stebalien:

We can have upgradability with one contract per SP. Basically, use DELEGATECALL.
Every SP has a separate contract with separate state.
Each contract uses DELEGATECALL to use the code from the shared contract.
That's how these things usually work anyways as deploying this way costs much less (one large contracts, many small ones). The upgrade path can also be opt-in (i.e., SPs can decide if/when they want to upgrade, all though the upgrade targets can be restricted).

Still not sure if this is worth supporting for our first version but worth investigating.

ZenGround0:
More on this.

We have settled on a singleton PDP service contract but this pattern is still useful. It might make sense to keep the contract upgradeable for a while as users test it out and then when it matures we can get rid of upgradeability.

PR #57, opened 2024-10-13
Feat/upgradeable

ZenGround0:
This addresses most of #23
We should probably also add the same upgradeability to the record keeper as a follow on.

Tested that upgrade works on calibration here: https://calibration.filfox.info/en/address/0x8e13508e629843bdcfc106c26989a3daa2d9ee74


contracts/src/PDPService.sol
```diff
@@ -84,13 +88,18 @@ contract PDPService {
     // proofset owner has exclusive permission to add and remove roots and delete the proof set
     mapping(uint256 => address) proofSetOwner;
     mapping(uint256 => address) proofSetProposedOwner;
-    
 
     // Methods
-    constructor(uint256 _challengeFinality) {
+    // Note: no constructor, use initialize()
```
anorth, line 93: Consider retaining a constructor that prevents initialisation of this contract directly (the initializer should be invoked on the proxy, not the impl).

```
/// @custom:oz-upgrades-unsafe-allow constructor
constructor() {
    _disableInitializers();
}
```

You may then need an `initializeForTest` initializer, or to use a proxy in the tests too.

anorth (approved):

ZenGround0:
Note to self -- tweaked the upgrade script to work for both contracts still need to test this on calibration net so potentially that has some minor bugs.

PR #58, opened 2024-10-14
Minor cleanups in testing

anorth:
I found this commit lying around that I think I intended to squash into a prior PR.


ZenGround0 (approved):
Thanks!

PR #60, opened 2024-10-15
Draft of proposed fees

ZenGround0:
Implementation of proposed PDP Fees


jennijuju:
PDP proof verification fees

ZenGround0:
The motivation behind the proof fee is to accrue value to the filecoin network 

The fee we're burning is small and won't accrue significant value without very widespread adoption.  In the event of widespread adoption its possible people will fork this contract to avoid the fee.  However we're optimistically experimenting with deploying a service that burns a fee.  The thinking is that we can potentially develop applications and great UX for this service that bring a lot of users and develop a network effect so that SPs are still better off using it even with a fee. 

People at filoz are also interested in seeing how this works out.  It might give useful insights we can incorporate in the design of the proposed work removing the batch balancer in favor of a similar proof fee (https://github.com/filecoin-project/FIPs/discussions/557)

ZenGround0:
This PR also contains a sybilFee which simply exists to prevent people from spamming proofsets in the same way miner creation has spammed miner actors.

irenegia:
@jennijuju 
> PDP proof verification fees
In this draft we have two fees, the proof fee that right now is though as "SP pays because using the network to verify an inclusion proof” and service fee "pay a small percentage of your earning to the network” . This may change  in the future, design work stil WIP

contracts/src/Fees.sol
```diff
@@ -0,0 +1,60 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.20;
+
+library PDPFees {
+
+    // Estimate of USD price in FIL
+    uint256 constant FIL_USD_NUM = 10;
+    uint256 constant FIL_USD_DENOM = 35;
```
anorth, line 8: Numerator/denominator seem backwards. If these really are the intended numbers, please explain in a comment (perhaps do so anyway).

contracts/src/Fees.sol
```diff
@@ -0,0 +1,60 @@
...
+        // [(Monthly revenue in USD) * (FIL / USD) * size * duration] / 
+        // [Leaves per TiB * Epochs per month * 1%]
+        uint256 numerator = ONE_FIL * MONTHLY_TIB_REVENUE_USD_NUM * FIL_USD_NUM * size * duration;
+        uint256 denominator = MONTHLY_TIB_REVENUE_USD_DENOM * FIL_USD_DENOM * LEAVES_PER_TIB * EPOCHS_PER_MONTH * ONE_PERCENT_DENOM;
+        return numerator / denominator;
+    }
+
+    // Returns the proof fee for a given challenge count
+    function proofFee(uint256 challengeCount) internal view returns (uint256) {
+        uint256 gasPrice;
```
anorth, line 44: Naming this "gas" really confuses me, for similar reasons to usage of the term "gas" in the spec for the batch balancer. This isn't gas. I'm not necessarily opposed to referencing the base fee to derive the proof fee (well, actually I am but agree to disagree) but please don't call the fee "gas".

Documentation would help a lot. What I think this is trying to do is "set the proof fee approximately equal to the gas cost of validating the proof (with a floor if the base fee is low). The gas cost will still be paid, so this is effectively doubling the computational cost of proof verification".

contracts/src/PDPVerifier.sol
```diff
@@ -229,7 +230,15 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
     // A proof set is created empty, with no roots. Creation yields a proof set ID
     // for referring to the proof set later.
     // Sender of create message is proof set owner.
-    function createProofSet(address listenerAddr) public returns (uint256) {
+    function createProofSet(address listenerAddr) public payable returns (uint256) {
+        uint256 sybilFee = PDPFees.sybilFee();
+        require(msg.value >= sybilFee, "sybil fee not met");
```
anorth, line 235: This check is already performed in `burnFee`. It's confusing to have it here but not in `provePosession`.

contracts/src/PDPVerifier.sol
```diff
@@ -344,16 +353,22 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
...
         require(proofs.length > 0, "empty proof");
-        uint256 seed = drawChallengeSeed(setId);
 
+        // Calculate and burn the proof fee
+        uint256 proofFee = PDPFees.proofFee(proofs.length);
+        burnFee(proofFee);
+        if (msg.value > proofFee) {
+            // Return the overpayment
+            payable(msg.sender).transfer(msg.value - proofFee);
+        }
```
anorth, line 367: Can you extract the "burn and return change" logic into a function for re-use? Directly into `burnFee` is probably enough.

PR #63, opened 2024-10-18
Enforce challenge count in simple pdp service

ZenGround0:
As part of requirements we want to allow on chain enforcement of challenge count during proving without baking assumptions directly in the pdp verifier.

So we want to put this in the simple pdp service.  To do this well wrt testing we want to refactor simple pdp service a bit.  This PR extracts record keeper functionality to a base contract that can be shared by simple pdp service and a testing contract that forgoes the complexity of simple pdp service.  

This PR also introduces functions on simple pdp service that make discovery of faults easy for off chain observers.

### DIFF
--- a/contracts/src/SimplePDPService.sol
+++ b/contracts/src/SimplePDPService.sol
@@ -6,8 +6,6 @@ import "../lib/openzeppelin-contracts-upgradeable/contracts/proxy/utils/Initiali
 import "../lib/openzeppelin-contracts-upgradeable/contracts/proxy/utils/UUPSUpgradeable.sol";
 import "../lib/openzeppelin-contracts-upgradeable/contracts/access/OwnableUpgradeable.sol";
 
-
-
 // PDPRecordKeeper tracks PDP operations.  It is used as a base contract for PDPListeners
 // in order to give users the capability to consume events async.
 contract PDPRecordKeeper {
@@ -35,7 +33,7 @@ contract PDPRecordKeeper {
     // Mapping to store events for each proof set
     mapping(uint256 => EventRecord[]) public proofSetEvents;
 
-    function receiveProofSetEvent(uint256 proofSetId, OperationType operationType, bytes memory extraData ) internal {
+    function receiveProofSetEvent(uint256 proofSetId, OperationType operationType, bytes memory extraData ) internal returns(uint256) {
         uint64 epoch = uint64(block.number);
         EventRecord memory newRecord = EventRecord({
             epoch: epoch,
@@ -45,6 +43,7 @@ contract PDPRecordKeeper {
         });
         proofSetEvents[proofSetId].push(newRecord);
         emit RecordAdded(proofSetId, epoch, operationType);
+        return proofSetEvents[proofSetId].length - 1;
     }
 
     // Function to get the number of events for a proof set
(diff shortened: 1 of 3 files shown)


ZenGround0:
@anorth please take a look at this next week and leave review / PR improvements.

contracts/src/SimplePDPService.sol
```diff
@@ -69,50 +128,51 @@ require(_pdpServiceAddress != address(0), "PDP service address cannot be zero");
...
 
     function rootsScheduledRemove(uint256 proofSetId, uint256[] memory rootIds) external onlyPDPVerifier {
         receiveProofSetEvent(proofSetId, OperationType.REMOVE_SCHEDULED, abi.encode(rootIds));
     }
 
+    // possession proven checks for correct challenge count and reverts if too low
+    // it also checks that proofs are not late and emits a fault record if so
     function posessionProven(uint256 proofSetId, uint256 challengedLeafCount, uint256 seed, uint256 challengeCount) external onlyPDPVerifier {
         receiveProofSetEvent(proofSetId, OperationType.PROVE_POSSESSION, abi.encode(challengedLeafCount, seed, challengeCount));
+        emit Debug("Here we go", 0);
```
anorth, line 145: Remove the debug events when you come back to clean up.

contracts/src/SimplePDPService.sol
```diff
@@ -69,50 +128,51 @@ require(_pdpServiceAddress != address(0), "PDP service address cannot be zero");
...
-        return proofSetEvents[proofSetId];
+        // check for unsubmitted proof 
+        if (!provenThisPeriod[proofSetId]) {
+            uint256 periodsSkipped = 1;
+            if (provingDeadlines[proofSetId] < block.number) {
+                periodsSkipped = 1 + ((block.number - provingDeadlines[proofSetId]) / getMaxProvingPeriod());
+            }
+            emit FaultRecord(FaultType.SKIPPED, periodsSkipped);
+        }
+        provenThisPeriod[proofSetId] = false;
```
anorth, line 176: This appears to be missing re-setting `provingDeadlines[proofSetId]` to a future epoch.

There's a policy decision to be made about which future epoch. I reckon that if it _was_ proven this period, we should add the full proving period to the deadline (so it stays the same time of day). So proving early doesn't lead to disadvantage of needing to prove more frequently over the long term.

But if this period wasn't proven, we probably want to set the deadline based on current epoch, like when the first root is added.

contracts/test/SimplePDPService.t.sol
```diff
@@ -78,4 +78,100 @@ contract SimplePDPServiceTest is Test {
...
+        uint256 proofSetId = 1;
+        uint256 challengedLeafCount = 100;
+        uint256 seed = 12345;
+        uint256 challengeCount = 5;
+
+        // Set up the proving deadline
+        pdpService.rootsAdded(proofSetId, 0, new PDPVerifier.RootData[](0));
+        // Warp to just before the deadline
+        vm.warp(block.number + pdpService.getMaxProvingPeriod() - 1);
+        pdpService.posessionProven(proofSetId, challengedLeafCount, seed, challengeCount);
```
anorth, line 115: I believe the proof should also be accepted _on_ the deadline, so exercise that here.

Issue #66, opened 2024-10-22
updating proving deadline


irenegia:
https://github.com/FILCAT/pdp/blob/14fd1c9225471aab4a517abdaf4791c8d9963fd4/contracts/src/SimplePDPService.sol#L132C59-L132C78

@ZenGround0 , this is the line that initialise the proving deadline of a proofset, but we can not find where the proving deadline is updated.

cc @lucaniz

anorth:
Yes, I think this is the same thing I noted in https://github.com/FILCAT/pdp/pull/63#discussion_r1807935410. The PR was merged "to be reviewed" so I expect this to be fixed up when Zen is back online.

Issue #65, opened 2024-10-22
behaviour of `nextProving period` in simplePDP


irenegia:
```
if (!provenThisPeriod[proofSetId]) {
            uint256 periodsSkipped = 1;
            if (provingDeadlines[proofSetId] < block.number) {
                periodsSkipped = 1 + ((block.number - provingDeadlines[proofSetId]) / getMaxProvingPeriod());
            }
            emit FaultRecord(FaultType.SKIPPED, periodsSkipped);

```

@ZenGround0, I believe that when `nextProvingPeriod` is called before a proof is received, the DPD service should revert the removals done by the PDPverifier function and do nothing else. All the faults accounting should be done in the pdp service `possesionProven` function

cc @lucaniz

https://github.com/FILCAT/pdp/blob/14fd1c9225471aab4a517abdaf4791c8d9963fd4/contracts/src/SimplePDPService.sol#L169

anorth:
Yes, I think this is the same thing I noted in https://github.com/FILCAT/pdp/pull/63#discussion_r1807935410. The PR was merged "to be reviewed" so I expect this to be fixed up when Zen is back online.

anorth:
I chatted with @irenegia and team about this. They're attempting to avoid a caller being marked faulty potentially unnecessarily (they made a mistake?), before the deadline has elapsed. One possible change would be to disallow this method before the deadline has passed.

This would add a little complexity and reduce freedom. At the moment I don't think it's quite worth it, but it's something to consider.

PR #67, opened 2024-10-23
Simplify fault design

ZenGround0:


contracts/src/SimplePDPService.sol
```diff
@@ -170,9 +164,14 @@ contract SimplePDPService is PDPListener, PDPRecordKeeper, Initializable, UUPSUp
             uint256 periodsSkipped = 1;
             if (provingDeadlines[proofSetId] < block.number) {
                 periodsSkipped = 1 + ((block.number - provingDeadlines[proofSetId]) / getMaxProvingPeriod());
+                provingDeadlines[proofSetId] = block.number + getMaxProvingPeriod(); // reset deadline
```
magik6k, line 167: Shouldn't that also be `provingDeadlines[proofSetId] = provingDeadlines[proofSetId] + getMaxProvingPeriod();` to keep the deadline in a fixed window?

The current way the proving window will keep sliding backwards as I have to compute and submit the proof some time before the deadline

ZenGround0: SGTM but we need to catch up multiple windows if we missed multiple otherwise you need to call this a bunch.  So you'll need to add periodsSkipped I think

ZenGround0: Can't PR now but fee lfree to push

anorth: Some [discussion about policy options](https://github.com/FILCAT/pdp/pull/63#discussion_r1807935410) here, it wants a bit of thought.

anorth (commented):
This would benefit from some tests demonstrating the behaviour, especially around when the proving deadline moves.

ZenGround0:
Because things were getting too complicated I did a small redesign [here](https://docs.google.com/document/d/1D3YUcva--MVrFNFbPwH0zezIKaJEg1X64VI-rf-XKnE/edit?tab=t.0).  This should now close 
#71 #66 and #65

contracts/src/SimplePDPService.sol
```diff
@@ -142,37 +134,51 @@ contract SimplePDPService is PDPListener, PDPRecordKeeper, Initializable, UUPSUp
...
+        uint256 periodsSkipped = (block.number - provingDeadlines[proofSetId]) / getMaxProvingPeriod();
+        uint256 faultPeriods = periodsSkipped;
+        if (!provenThisPeriod[proofSetId]) { 
+            // include previous unproven period 
+            faultPeriods += 1;
+        }
+        if (faultPeriods > 0) {
+            emit FaultRecord(faultPeriods);
         }
+        provingDeadlines[proofSetId] = provingDeadlines[proofSetId] + getMaxProvingPeriod()*periodsSkipped;
```
anorth, line 181: Nit: I think this code could also handle the on-time case, with something like `periodsSkipped = max(0, (block.number - provingDeadlines[proofSetId])) / getMaxProvingPeriod()`. The rest is basically duplicate.

contracts/src/SimplePDPService.sol
```diff
@@ -142,37 +134,51 @@ contract SimplePDPService is PDPListener, PDPRecordKeeper, Initializable, UUPSUp
...
-        // check for unsubmitted proof 
-        if (!provenThisPeriod[proofSetId]) {
-            uint256 periodsSkipped = 1;
-            if (provingDeadlines[proofSetId] < block.number) {
-                periodsSkipped = 1 + ((block.number - provingDeadlines[proofSetId]) / getMaxProvingPeriod());
+        // Noop when proving period not yet open
+        // Can only get here if calling nextProvingPeriod multiple times within the same proving period
+        uint256 prevDeadline = provingDeadlines[proofSetId] - getMaxProvingPeriod();
+        if (block.number <= prevDeadline) {
+            return;
```
anorth, line 160: I think this needs to revert. As is, a provider could grind challenges by calling `verifier.nextProvingPeriod` once each epoch after a successful proof until the deadline is reached.

In general, I think any call to this method must either (1) advance the proving period (possibly with fault), or (2) revert.

anorth: This makes me think that, in the PDP Verifier contract, the method name `nextProvingPeriod`  is not quite right. That contract doesn't know anything about proving periods. It could instead be `newChallenge` or something. However since that's the method actually invoked by SPs, I can see the motivation for a name that represents their expected workflow.

contracts/test/SimplePDPService.t.sol
```diff
@@ -93,85 +93,152 @@ contract SimplePDPServiceTest is Test {
...
     }
 
-    function testPosessionProvenLate() public {
-        uint256 proofSetId = 1;
-        uint256 challengedLeafCount = 100;
-        uint256 seed = 12345;
-        uint256 challengeCount = 5;
+    function testNextProvingPeriodCalledLastMinuteOK() public {
+        pdpService.rootsAdded(proofSetId, 0, new PDPVerifier.RootData[](0));
+        vm.roll(block.number + pdpService.getMaxProvingPeriod() - 1);
```
anorth, line 123: By my reading, this should also succeed one epoch later (and same below).

Could you add a test for the first epoch at which the proof is too late?

anorth (commented):
I think this is very close.

The grinding that I see remaining could only be exercised with integration tests, which we lack. I suggest following up to add some integration tests, with some of them explicitly constructed from an "SP trying to cheat" frame.

contracts/test/SimplePDPService.t.sol
ZenGround0: Thanks this helped me find a bug

contracts/src/SimplePDPService.sol
ZenGround0: Thanks for the catch on the revert.

Yeah I'm still not happy with the name but there is some sort of proving period thing that the pdp verifier knows about.  Namely when batched adds and removes get added into the proofset.  I think this name is probably ok for now even though pdp service has a modified idea of proving period

test/PDPVerifier.t.sol
```diff
@@ -23,10 +23,7 @@ contract PDPVerifierProofSetCreateDeleteTest is Test {
         );
         MyERC1967Proxy proxy = new MyERC1967Proxy(address(pdpVerifierImpl), initializeData);
         pdpVerifier = PDPVerifier(address(proxy));
-        SimplePDPService listenerImpl = new SimplePDPService();
-        initializeData = abi.encodeWithSelector(SimplePDPService.initialize.selector, address(pdpVerifier));
-        MyERC1967Proxy listenerProxy = new MyERC1967Proxy(address(listenerImpl), initializeData);
-        listener = SimplePDPService(address(listenerProxy));
+        listener = new TestingRecordKeeperService();
```
anorth, line 26: ❤️

anorth (approved):

PR #76, opened 2024-11-19
Fix Foundry project structure

aarshkshah1992:
Ensure project structure follows https://book.getfoundry.sh/projects/project-layout for easier CLI, CI and more. Also add build artefacts to gitignore.


.gitignore
```diff
@@ -1,2 +1,19 @@
 # Visual Studio Code local configuration
 .vscode
```
ZenGround0, line 2: I think you can remove lines 1 and 2 now that you've added 4 and 5

Makefile
```diff
@@ -1,23 +1,35 @@
...
-	cd contracts && ../tools/deploy-calibnet.sh
+	./tools/deploy-calibnet.sh
 
+.PHONY: deploy-devnet
 deploy-devnet:
-	cd contracts && ../tools/deploy-devnet.sh
+	./tools/deploy-devnet.sh
 
+# Test utilities
+.PHONY: testBurnFee
```
ZenGround0, line 33: You can remove this target and the associated script now that we've completed testing this

ZenGround0 (commented):
Its ok you've removed the out/ directory since that was not the best.  But FYI we're going to need some better way to provide @magik6k with the abi info that curio needs, maybe a Makefile target?  We can sort that out after this PR though.

Makefile
aarshkshah1992: Done.

aarshkshah1992:
@ZenGround0 Addressed your review and CI is now green. Please can you approve ?

PR #77, opened 2024-11-21
Challenge window restriction

ZenGround0:
Enforce the proof of possession challenge epoch to be within a 30 minute challenge window at the end of the proving period.  

The motivation is 
1) This forces 24 hour proof of access rather than a single period of access the duration of challengeFinality epochs every 48 hours fixing a problem pointed out by @irenegia and @lucaniz

2) Because 30 minutes is too restrictive for unsealing this adds a time guarantee that PDP SPs truly have an unsealed copy and are not unsealing for proofs fixing a problem pointed out by @magik6k 

To do this with the existing architecture this PR adds a parameter to the NextProvingPeriod method on the verifier allowing the user to pass in the challengeEpoch.  And it adds the parameter to the listener interface.  Now the PDPService can enforce its own requirements on the challengeEpoch.

### DIFF
--- a/src/SimplePDPService.sol
+++ b/src/SimplePDPService.sol
@@ -105,26 +105,40 @@ contract SimplePDPService is PDPListener, PDPRecordKeeper, Initializable, UUPSUp
         return 2880;
     }
 
-    // Number of epochs at the end of a provoing period during which a 
+    // Number of epochs at the end of a proving period during which a 
     // proof of possession can be submitted
     function challengeWindow() public pure returns (uint256) {
         return 60;
     }
 
-    // The start of the next challenge window
-    // Useful for querying before nextProvingPeriod to determine challengeEpoch
-    function nextChallengeWindowStart(uint256 setId) public view returns (uint256) {
+    // The start of the challenge window for the current proving period
+    function thisChallengeWindowStart(uint256 setId) public view returns (uint256) {
         if (provingDeadlines[setId] == 0) {
-            revert("Proving period not yet open");
+            revert("Proving not yet started");
         }
+
         uint256 periodsSkipped;
         // Proving period is open 0 skipped periods
         if (block.number <= provingDeadlines[setId]) {
             periodsSkipped = 0;
         } else { // Proving period has closed possibly some skipped periods
-            periodsSkipped = (block.number - (provingDeadlines[setId] + 1)) / getMaxProvingPeriod();
+            periodsSkipped = 1 + (block.number - (provingDeadlines[setId] + 1)) / getMaxProvingPeriod();
         }
-        return provingDeadlines[setId] + getMaxProvingPeriod()*(periodsSkipped+1) - challengeWindow();
+        return provingDeadlines[setId] + periodsSkipped*getMaxProvingPeriod() - challengeWindow();
+    }
+
+    // The start of the NEXT OPEN proving period's challenge window
+    // Useful for querying before nextProvingPeriod to determine challengeEpoch to submit for nextProvingPeriod
+    function nextChallengeWindowStart(uint256 setId) public view returns (uint256) {
+        if (provingDeadlines[setId] == 0) {
+            revert("Proving not yet started");
+        }
+        // If the current period is open this is the next period's challenge window
+        if (block.number <= provingDeadlines[setId]) {
+            return thisChallengeWindowStart(setId) + getMaxProvingPeriod();
+        }
+        // If the current period is not yet open this is the current period's challenge window
+        return thisChallengeWindowStart(setId);
     }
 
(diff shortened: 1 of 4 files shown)


src/SimplePDPService.sol
```diff
@@ -105,6 +105,28 @@ contract SimplePDPService is PDPListener, PDPRecordKeeper, Initializable, UUPSUp
...
+    // Number of epochs at the end of a provoing period during which a 
+    // proof of possession can be submitted
+    function challengeWindow() public pure returns (uint256) {
+        return 60;
+    }
+
+    // The start of the next challenge window
+    // Useful for querying before nextProvingPeriod to determine challengeEpoch
+    function nextChallengeWindowStart(uint256 setId) public view returns (uint256) {
+        if (provingDeadlines[setId] == 0) {
```
lucaniz, line 117: Does this If apply only when the proofSet is created (i.e. provingDeadlines[setId] is set to 0 when the proofSet is created?

src/SimplePDPService.sol
```diff
@@ -164,9 +186,15 @@ contract SimplePDPService is PDPListener, PDPRecordKeeper, Initializable, UUPSUp
...
         if (block.number <= provingDeadlines[proofSetId]) {
             periodsSkipped = 0;
-        } else { // Proving period has closed possible some skipped periods
+        } else { // Proving period has closed possibly some skipped periods
             periodsSkipped = (block.number - (provingDeadlines[proofSetId] + 1)) / getMaxProvingPeriod();
         }
+        // ensure next challenge epoch falls within the next challenge window.
+        // The next challenge window immediately precedes the next deadline
+        uint256 nextDeadline = provingDeadlines[proofSetId] + getMaxProvingPeriod()*(periodsSkipped+1);
+        if (challengeEpoch < nextDeadline - challengeWindow() || challengeEpoch > nextDeadline) {
```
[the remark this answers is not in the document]
ZenGround0: We want to maintain separation of concerns between the PDPVerifier and the SimplePDPService.  The way it is here we can implement a new PDPService on top of the same PDPVerifier contract with a shorter challenge window, a longer challenge window or a different structure.

src/SimplePDPService.sol
ZenGround0: Yes.  In solidity variables are initialized to their zero value.  There should be no way for the code to assign 0 to provingDeadlines[setId] after the first assignment.

src/SimplePDPService.sol
```diff
@@ -105,6 +105,28 @@ contract SimplePDPService is PDPListener, PDPRecordKeeper, Initializable, UUPSUp
...
+            revert("Proving period not yet open");
+        }
+        uint256 periodsSkipped;
+        // Proving period is open 0 skipped periods
+        if (block.number <= provingDeadlines[setId]) {
+            periodsSkipped = 0;
+        } else { // Proving period has closed possibly some skipped periods
+            periodsSkipped = (block.number - (provingDeadlines[setId] + 1)) / getMaxProvingPeriod();
+        }
+        return provingDeadlines[setId] + getMaxProvingPeriod()*(periodsSkipped+1) - challengeWindow();
```
anorth, line 127: Please add tests for this method. Arithmetic like this is is tricky to mentally verify.

anorth (commented):
LGTM after tests

src/SimplePDPService.sol
```diff
@@ -105,6 +105,28 @@ contract SimplePDPService is PDPListener, PDPRecordKeeper, Initializable, UUPSUp
         return 2880;
     }
 
+    // Number of epochs at the end of a provoing period during which a
```
aarshkshah1992, line 108: proving*

src/SimplePDPService.sol
aarshkshah1992: @ZenGround0 Can you explain what we're doing here ? Where's this arithmetic coming from ?

ZenGround0: Ok I refactored this to be a lot less confusing.  The tricky thing is that if the current epoch falls before the latest marked deadline we want to roll to the *next* period's challenge window.  But if we've faulted, are late and the current epoch is well after the latest marked deadline we want to roll to the *current* period's window.  This is all now a lot more explicit.

aarshkshah1992: Yeah this all makes sense now 👍

aarshkshah1992:
@ZenGround0 LGTM modulo the changes we discussed in the call today

1) Update `ChallengeDelay` to 150 epochs
2) Ensure SP calls `nextProvingPeriod` as part of initialisation -> then adds roots.

aarshkshah1992 (approved):

PR #78, opened 2024-11-22
Update interface

### LINKED ISSUE #56: PDP Methods should take in extra data
The PDP Listener interface should allow an extra data param to be threaded through from the caller.  This means we should add extra data params into the main PDP interface to allow caller to talk directly to the listener.  This lets us use the safer, cleaner listener pattern but still allow for protocol extensions in base PDP that allow offchain auth and other extensions.  The canonical example is an application that allows PDP client to fully determine the allowlist of valid cids to be added to proofset.

1. Without extra data: in this case the PDP client must go to the application contract directly and store the valid cids that are allowed to be added to the proofset in the contract state before the PDP SP can call PDP addroots.
2. With extra data: signature is sent to PDP SP who then includes it and passes to listener during addRoots call

ZenGround0:
Closes #56 giving PDPService implementations the option to do client-signs-sp-submits protocols as in today's builtin storage market.

Will take out of draft once the challenge window PR upon which this is based is merged

### DIFF
--- a/src/PDPVerifier.sol
+++ b/src/PDPVerifier.sol
@@ -15,6 +15,7 @@ interface PDPListener {
     function proofSetDeleted(uint256 proofSetId, uint256 deletedLeafCount, bytes calldata extraData) external;
     function rootsAdded(uint256 proofSetId, uint256 firstAdded, PDPVerifier.RootData[] memory rootData, bytes calldata extraData) external;
     function rootsScheduledRemove(uint256 proofSetId, uint256[] memory rootIds, bytes calldata extraData) external;
+    // Note: extraData not included as proving messages conceptually always originate from the SP 
     function posessionProven(uint256 proofSetId, uint256 challengedLeafCount, uint256 seed, uint256 challengeCount) external;
     function nextProvingPeriod(uint256 proofSetId, uint256 challengeEpoch, uint256 leafCount, bytes calldata extraData) external;
 }
@@ -26,7 +27,7 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
     uint256 public constant MAX_ROOT_SIZE = 1 << 50;
     uint256 public constant MAX_ENQUEUED_REMOVALS = 2000;
     address public constant RANDOMNESS_PRECOMPILE = 0xfE00000000000000000000000000000000000006;
-    uint256 public constant EXTRA_DATA_MAX_SIZE = 512;
+    uint256 public constant EXTRA_DATA_MAX_SIZE = 2048;
 
     // Events
     event ProofSetCreated(uint256 indexed setId);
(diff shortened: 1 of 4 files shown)


src/PDPVerifier.sol
```diff
@@ -11,12 +11,12 @@ import "../lib/openzeppelin-contracts-upgradeable/contracts/access/OwnableUpgrad
...
 interface PDPListener {
-    function proofSetCreated(uint256 proofSetId, address creator) external;
-    function proofSetDeleted(uint256 proofSetId, uint256 deletedLeafCount) external;
-    function rootsAdded(uint256 proofSetId, uint256 firstAdded, PDPVerifier.RootData[] memory rootData) external;
-    function rootsScheduledRemove(uint256 proofSetId, uint256[] memory rootIds) external;
+    function proofSetCreated(uint256 proofSetId, address creator, bytes calldata extraData) external;
+    function proofSetDeleted(uint256 proofSetId, uint256 deletedLeafCount, bytes calldata extraData) external;
+    function rootsAdded(uint256 proofSetId, uint256 firstAdded, PDPVerifier.RootData[] memory rootData, bytes calldata extraData) external;
+    function rootsScheduledRemove(uint256 proofSetId, uint256[] memory rootIds, bytes calldata extraData) external;
     function posessionProven(uint256 proofSetId, uint256 challengedLeafCount, uint256 seed, uint256 challengeCount) external;
```
anorth, line 18: Out of curiosity, why omit the extra data on this method? It will seem odd to a reader, so perhaps you could document it.

src/PDPVerifier.sol
```diff
@@ -26,6 +26,7 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
     uint256 public constant MAX_ROOT_SIZE = 1 << 50;
     uint256 public constant MAX_ENQUEUED_REMOVALS = 2000;
     address public constant RANDOMNESS_PRECOMPILE = 0xfE00000000000000000000000000000000000006;
+    uint256 public constant EXTRA_DATA_MAX_SIZE = 512;
```
anorth, line 29: What's the justification for this number? Again, perhaps documents it. It seems rather limiting – what could go wrong if we chose a larger number?

anorth (approved):
I suggest adding at least one test that the extra data makes it through to the application/service, even though it looks straightforward and is largely constrained by types.

Otherwise LGTM

src/PDPVerifier.sol
ZenGround0: We could go bigger for sure, let's double it.  My thinking was 16 words are so much that its hard to think of protocols that are going to restricted by it but I think we get a little closer to the limits of PSD deal proposal so bigger is probably better.

aarshkshah1992 (approved):

PR #80, opened 2024-11-30
Proving Schedule interface

ZenGround0:
This makes it easier for curio or other SP tools to consume standard proving window meta data that is likely to be standard across listeners.  See integration into curio [here](https://github.com/filecoin-project/curio/pull/334)

### DIFF
--- a/src/IPDPProvingSchedule.sol
+++ b/src/IPDPProvingSchedule.sol
@@ -1,0 +1,23 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.20;
+
+/// @title IPDPProvingWindow
+/// @notice Interface for PDP Service SLA specifications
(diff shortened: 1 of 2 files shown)


src/IPDPProvingSchedule.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.20;
+
+/// @title IPDPProvingWindow
+/// @notice Interface for PDP Service SLA specifications
```
anorth, line 5: Perhaps note that this is specific to the SimplePDPService implemented here, and other services could have different models (e.g. without proving periods).

anorth (approved):

aarshkshah1992 (approved):

aarshkshah1992:
lgtm.

PR #82, opened 2024-12-04
feat: implement PDP V0 proof fees

aarshkshah1992:
Implements PDP proof fees as flushed out and documented [here](https://www.notion.so/filecoindev/Pricing-mechanism-for-PDPverifier-12adc41950c180ea9608cb419c369ba4).

The Proof fees are proportional to the number of challenges and size of the proof set being proved. They're burnt and are paid by the SP once every proving period when they call `provePossesion`.

However, if the gas fee for a call to `provePossesion` is >= 5 % of daily estimated storage reward for an SP, then no proof fee is charged.

Please see the linked doc for more details.


Makefile
```diff
@@ -13,12 +13,12 @@ all: build test
 # Build target
 .PHONY: build
 build:
-	forge build
+	forge build --via-ir
```
ZenGround0, line 16: Hmm where is this coming from?  Are we exceeding compiler's threshold for happiness if we don't do this?  Can we instead remove local variables from functions?

src/Log2.sol
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.20;
+
+/// @notice Gas-efficient library for log2 calculations
+library Log2 {
```
ZenGround0, line 5: nit: I think this belongs next to the BitOps library in bitops.sol

ZenGround0: Since we've already built and vetted other bit ops stuff it might reduce our code surface and make this simpler if you just use what we've already built and do

```solidity
return 255 - clz(value) 
```

Does that sound ok?

aarshkshah1992: Didn't realise this can be used for log2 as well. Will dig into it and replace.

src/PDPVerifier.sol
```diff
@@ -354,41 +354,74 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
...
+        burnFee(proofFee);
+        if (msg.value > proofFee) {
+            // Return the overpayment
+            payable(msg.sender).transfer(msg.value - proofFee);
+        }
+
         address listenerAddr = proofSetListener[setId];
         if (listenerAddr != address(0)) {
-            PDPListener(listenerAddr).posessionProven(setId, proofSetLeafCount[setId], seed, proofs.length);
+            PDPListener(listenerAddr).posessionProven(
```
ZenGround0, line 419: nit: without a linter enforcing a particular style these editor style refactors are noisy and probably best to avoid.  I would guess the style of the code is now less consistent since the style change here only shows up in places that you've edited.

src/PDPVerifier.sol
```diff
@@ -354,41 +354,74 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
...
+            bool ok = MerkleVerify.verify(
+                proofs[i].proof,
+                rootHash,
+                proofs[i].leaf,
+                root.offset
+            );
             require(ok, "proof did not verify");
         }
 
+        uint256 gasUsed = (initialGas - gasleft()) + (callDataSize * 1300);
```
ZenGround0, line 402: Good idea to do this above the listener as otherwise they could do all sorts of work to increase the measured gas and avoid paying fees.  Its worth putting a comment explaining this ordering here.

ZenGround0 (commented):
Looking good I'll do a more thorough review tomorrow.

Makefile
aarshkshah1992: @ZenGround0 Very tricky to get rid of this . We need the local variables in `provePossession` to calculate gas fee and refactoring the code to get rid of these makes the code hard to read.

But I looked around and it's not necessarily bad  to use this flag i.e. it's not a smell/bad practise and is common with non-trivial contracts.

src/Fees.sol
```diff
@@ -1,58 +1,66 @@
...
-        // [Leaves per TiB * Epochs per month * 1%]
-        uint256 numerator = ONE_FIL * MONTHLY_TIB_REVENUE_USD_NUM * FIL_USD_NUM * size * duration;
-        uint256 denominator = MONTHLY_TIB_REVENUE_USD_DENOM * FIL_USD_DENOM * LEAVES_PER_TIB * EPOCHS_PER_MONTH * ONE_PERCENT_DENOM;
-        return numerator / denominator;
-    }
+    // 0.1 FIL
+    uint256 constant SYBIL_FEE = ONE_FIL / 10;
+
+    // assume 1FIL = $5 for now -> we can change this to use an oracle in PDP V1
+    uint256 constant FIL_USD_PRICE = 5;
```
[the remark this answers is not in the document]
aarshkshah1992: @ZenGround0 

See discussion [here](https://www.notion.so/Pricing-mechanism-for-PDPverifier-12adc41950c180ea9608cb419c369ba4?d=151dc41950c180889134001c2a9d89b6&pvs=4#145dc41950c180a78fb5ee20ef9958ca)

1) We can either ship this now with this price for PDP V0 and revisit in PDP V1 etc once the on chain payments/FWS story becomes clearer

2) Change this to use an oracle right now that gives us a real time price feed and live with the complexity.

What are your thoughts ?

src/Log2.sol
aarshkshah1992: Done.

test/PDPVerifier.t.sol
```diff
@@ -600,28 +602,21 @@ contract PDPVerifierProofTest is Test, ProofBuilderHelper {
...
         PDPVerifier.Proof[] memory proofs = buildProofsForSingleton(setId, challengeCount, tree, leafCount);
 
-        // Calculate the correct proof fee
-        uint256 correctProofFee = PDPFees.proofFee(challengeCount);
-
         // Test 1: Sending less than the required fee
-        uint256 insufficientFee = correctProofFee - 1;
         vm.expectRevert("Incorrect fee amount");
-        pdpVerifier.provePossession{value: insufficientFee}(setId, proofs);
+        pdpVerifier.provePossession{value: 1}(setId, proofs);
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

test/PDPVerifier.t.sol
```diff
@@ -1348,7 +1348,11 @@ contract PDPVerifierE2ETest is Test, ProofBuilderHelper {
         PDPVerifier.Proof[] memory proofsPP1 = buildProofs(pdpVerifier, setId, 5, treesA, leafCountsA);
 
         vm.mockCall(pdpVerifier.RANDOMNESS_PRECOMPILE(), abi.encode(pdpVerifier.getNextChallengeEpoch(setId)), abi.encode(pdpVerifier.getNextChallengeEpoch(setId)));
-        pdpVerifier.provePossession{value: PDPFees.proofFee(proofsPP1.length)}(setId, proofsPP1);
+        console.log("hello");
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

src/PDPVerifier.sol
```diff
@@ -354,41 +354,74 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
...
-        }
+        uint256 initialGas = gasleft();
+        uint256 callDataSize = 0;
 
         uint256 seed = drawChallengeSeed(setId);
         uint256 leafCount = challengeRange[setId];
         uint256 sumTreeTop = 256 - BitOps.clz(nextRootId[setId]);
         for (uint64 i = 0; i < proofs.length; i++) {
+            // Calculate call data size
+            // 32 for the leaf + each element in the proof is 32 bytes
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

src/PDPVerifier.sol
```diff
@@ -354,41 +354,74 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
...
+            callDataSize += 32 + (proofs[i].proof.length * 32);
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

src/PDPVerifier.sol
aarshkshah1992: Done.

src/Log2.sol
aarshkshah1992: Yeah this was a cool tip. Thanks. I grok why it works.

src/Fees.sol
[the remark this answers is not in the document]
irenegia: recaping here that we decided to move forward implementing a call to a price oracle already in v0!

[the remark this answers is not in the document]
lucaniz: [not relevant since we are going for the price oracle, just putting this here for clarity: in the current case, if the fil price is higher than what is set up here, the fee is going to be higher than expected (i.e. if FIL gets to 10 and price is set to 5 in the contract, fee would be 2x the expected value)]

src/Fees.sol
```diff
@@ -1,58 +1,66 @@
...
+    // 2 USD/Tib/month is the current reward earned by Storage Providers
+    uint256 constant MONTHLY_TIB_STORAGE_REWARD_USD = 2;
+
+    uint256 constant DAILY_TIB_STORAGE_REWARD_ATTO_FIL =
+        (MONTHLY_TIB_STORAGE_REWARD_USD * 1e18 * ATTO_FIL) /
+            (30 * FIL_USD_PRICE);
+
+    // PROOF_PRICE is currently set to 1% of the daily reward
+    uint256 constant PROOF_PRICE_ATTO_FIL =
+        (1 * DAILY_TIB_STORAGE_REWARD_ATTO_FIL) / 100;
```
lucaniz, line 24: Note that this is aligned 100% with what happens in simple PDP service. 
Just want to point out that if anyone deploys his own PDPService contract (with different frequency) er have two options
- make clear that in this case one should also update the proofFee definition accordingly
- define proofFee in terms of MaxProvingPeriod (i.e. Proof_Price_Atto_Fil = x*epochReward*getMaxProvingPeriod())

If this does not happen, in case of changing the proof frequency the SP is paying the 1% of the daily reward each time he posts a proof. 

Note that in the second suggested option we have as a prerequisite the access to getMaxProvingPeriod(), which "rules out" policies which are not based on proving with some frequency (for instance: prove upon client request, no fixed schedule).

src/Fees.sol
```diff
@@ -1,58 +1,66 @@
...
+
+    // 5% of daily reward
+    uint256 constant FIVE_PERCENT_DAILY_REWARD_ATTO_FIL =
+        (DAILY_TIB_STORAGE_REWARD_ATTO_FIL * 5) / 100;
+
```
lucaniz, line 29: Same comment as above holds here as well. 

And also for 4% daily reward.

src/Fees.sol
lucaniz: @aarshkshah1992 I think what we call Proof_price here is actually ProofFee in the doc https://www.notion.so/filecoindev/Pricing-mechanism-for-PDPverifier-12adc41950c180ea9608cb419c369ba4

src/Fees.sol
```diff
@@ -1,58 +1,66 @@
...
-        uint256 gasPrice;
-        if (block.basefee > ONE_NANO_FIL) {
-            gasPrice = block.basefee;
+        if (estimatedGasFee >= FIVE_PERCENT_DAILY_REWARD_ATTO_FIL) {
+            return 0; // No proof fee if gas fee is above 5% of the estimated reward
+        } else if (estimatedGasFee >= FOUR_PERCENT_DAILY_REWARD_ATTO_FIL) {
+            return FIVE_PERCENT_DAILY_REWARD_ATTO_FIL - estimatedGasFee; // Partial discount on proof fee
         } else {
-            gasPrice = ONE_NANO_FIL;
+            uint256 calculatedProofFee = (PROOF_PRICE_ATTO_FIL *
```
lucaniz, line 55: I think we are calculating ProofFee the wrong way. 

ProofPrice is supposed to be different than the one calculated above
See https://www.notion.so/filecoindev/Pricing-mechanism-for-PDPverifier-12adc41950c180ea9608cb419c369ba4?pvs=4#145dc41950c180f5be05f413ccf3b644

cc @irenegia

src/Fees.sol
lucaniz: See comment on Proof_Price below. the formula here is the formula of ProofFee wrt requirement 2 here 
https://www.notion.so/filecoindev/Pricing-mechanism-for-PDPverifier-12adc41950c180ea9608cb419c369ba4?pvs=4#13adc41950c180709d5ed5142aee845d

aarshkshah1992: @lucaniz Yeah renamed this variable to "proof fee" which is 1% of DAILY REWARD.

aarshkshah1992: @lucaniz @irenegia I am a bit confused on the right naming. Is 1% of the daily reward the "proof fee" or the "proof price" ? To me, "price" is more intuitive like price of 1 banana and then to get the "total proof fee" I scale it by proof size. Lemme know what you'll think.

aarshkshah1992: I think "proof price" and "proof cost" is a good naming convention here. But will defer to you'll.

src/Fees.sol
aarshkshah1992: Please see https://github.com/FilOzone/pdp/pull/82#discussion_r1871604208.

Makefile
aarshkshah1992: Actually fixing this.

aarshkshah1992: Really slows down compilation.

aarshkshah1992:
Raising a new PR for this as we've changed the price formulation.

ZenGround0:
@aarshkshah1992 what's the deal with checking in the node modules but not the solidity libraries?  When you cleaned up the repo you seemed to be keeping dependencies out.  Is there a good reason to keep some dependencies and not others?

aarshkshah1992:
@ZenGround0 Removed the node_modules/. We will install them as part of CI.

Unfortunately, as per Pyth docs -> https://docs.pyth.network/price-feeds/use-real-time-data/evm

we need to use node for Pyth. However, no biggie. Just run `make install`  locally and it will all be taken care of you.

aarshkshah1992:
@ZenGround0 Please don't review it yet (I mean your call). Wanna go through all of the code once again and convince myself it's correct (also wanna test on Calib).

src/Fees.sol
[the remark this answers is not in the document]
usdfil: Have you considered using a stablecoin as payment for PDP? We'd love to work with you to figure out how to make this happen!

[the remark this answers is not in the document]
jennijuju: @usdfil we plan build a payment pathway with fungible token support - but that will be the next milestone & will be outside of the proof contract itself (out of the scope of this PR). It is better to follow #fil-pdp, and we would like to get your support when we start working on payments.

aarshkshah1992:
@ZenGround0 This is now ready for review.

src/Fees.sol
```diff
@@ -1,58 +1,69 @@
...
-        // [Leaves per TiB * Epochs per month * 1%]
-        uint256 numerator = ONE_FIL * MONTHLY_TIB_REVENUE_USD_NUM * FIL_USD_NUM * size * duration;
-        uint256 denominator = MONTHLY_TIB_REVENUE_USD_DENOM * FIL_USD_DENOM * LEAVES_PER_TIB * EPOCHS_PER_MONTH * ONE_PERCENT_DENOM;
-        return numerator / denominator;
-    }
+    // 0.1 FIL
+    uint256 constant SYBIL_FEE = FIL_TO_ATTO_FIL / 10;
+
+    // 2 USD/Tib/month is the current reward earned by Storage Providers
+    uint256 constant MONTHLY_TIB_STORAGE_REWARD_USD = 2;
```
lucaniz, line 15: minor (feel free to ignore): I'd personally avoid the word "reward" to avoid misunderstading withblock reward. Maybe income? idk...

src/Fees.sol
```diff
@@ -1,58 +1,69 @@
...
+    // 1% of reward per period
+    uint256 constant PROOF_FEE_PERCENTAGE = 1;
+    // 4% of reward per period for gas limit left bound
+    uint256 constant GAS_LIMIT_LEFT_PERCENTAGE = 4;
+    // 5% of reward per period for gas limit right bound
+    uint256 constant GAS_LIMIT_RIGHT_PERCENTAGE = 5;
+    uint256 constant USD_DECIMALS = 1e18;
+
+    // 1 TiB in bytes (2^40)
+    uint256 constant TIB_IN_BYTES = 2 ** 40;
```
lucaniz, line 25: minor (again, feel free to ignore): i like more bytes_in_1TiB, but I guess it's personal taste

src/PDPVerifier.sol
```diff
@@ -174,6 +182,12 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         return nextChallengeEpoch[setId];
     }
 
+    // Returns the listener address for a proof set
+    function getProofSetListener(uint256 setId) public view returns (address) {
```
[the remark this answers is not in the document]
ZenGround0: I see you merged main into this branch so it makes some sense.  Weird that github UI sees this as a diff though seems almost like a github bug?  Probably just advanced git edge cases.

src/PDPVerifier.sol
```diff
@@ -584,4 +631,20 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
...
+            86400
+        );
+        require(priceData.price > 0, "failed to validate: price must be greater than 0");
+
+        // The Pyth price data represents USD per FIL as priceData.price * 10**(priceData.expo)
+        emit FILUSDPrice(priceData.price, priceData.expo);
+
+        // Return the price and exponent representing USD per FIL
+        return (uint64(priceData.price), priceData.expo);
+    }
```
ZenGround0, line 649: Thanks for getting this all so clean.  It looks great.

ZenGround0 (commented):
Reviewed Verifier, moving onto Fees, then tests.

src/PDPVerifier.sol
```diff
@@ -106,6 +113,7 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
     // proofset owner has exclusive permission to add and remove roots and delete the proof set
     mapping(uint256 => address) proofSetOwner;
     mapping(uint256 => address) proofSetProposedOwner;
+    mapping(uint256 => uint256) proofSetLastProvenEpoch;
```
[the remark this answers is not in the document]
ZenGround0: One thing this variable is missing here is initialization.  As it is the very first call of provePossession will take the interval since last proving to be the current filecoin block number - 0, i.e. 4 years which will cause the fee to max out.  Two approaches both seem fine to me
1. Initialize the variable to current block time upon first add.  Then I will change it to do this initialization upon first call to nextProvingPeriod in #86 
2. When proofSetLastProvenEpoch is unset we say there has been no time since last proof and the fee is 0 for the first call.

Maybe consult with @irenegia to see if there is any preference from fee design standpoint.

In either case you should have a constant NO_PROEN_EPOCH = 0.  When all leaves are removed from the proofset we should set proofSetLastProvenEpoch = NO_PROVEN_EPOCH.  This will handle the case where a proof set starts up deletes everything and starts up again.

src/Fees.sol
```diff
@@ -1,58 +1,69 @@
...
+        uint256 nProofEpochs
+    ) internal pure returns (uint256) {
+        require(estimatedGasFee > 0, "failed to validate: estimated gas fee must be greater than 0");
+        require(filUsdPrice > 0, "failed to validate: AttoFIL price must be greater than 0");
+        require(rawSize > 0, "failed to validate: raw size must be greater than 0");
+
+        // Calculate reward per epoch per byte (in AttoFIL)
+        uint256 rewardPerEpochPerByte;
+        if (filUsdPriceExpo >= 0) {
+            rewardPerEpochPerByte = (MONTHLY_TIB_STORAGE_REWARD_USD * FIL_TO_ATTO_FIL) / (TIB_IN_BYTES * EPOCHS_PER_MONTH * filUsdPrice * (10 ** uint32(filUsdPriceExpo)));
```
ZenGround0, line 44: Does pyth oracale have any notion of precision?  The way things are structured here we appear to lose precision exponentially as price increases.  Assuming filUsdPrice is an integer that needs not precision adjustments as we use it here and the pyth oracle can give us FIL price of $1,$2,$3,...,$9, $10, $20, $30,... $100, $200, $300.  If this is the best we can do it is massively better than just estimating with a constant. But ideally we could get constant precision for any price.

src/Fees.sol
```diff
@@ -1,58 +1,69 @@
...
+        } else {
+            rewardPerEpochPerByte = (MONTHLY_TIB_STORAGE_REWARD_USD * FIL_TO_ATTO_FIL * (10 ** uint32(-filUsdPriceExpo))) /
+            (TIB_IN_BYTES * EPOCHS_PER_MONTH * filUsdPrice);
+        }
+
+        // Calculate total reward for the proving period
+        uint256 rewardPerPeriod = rewardPerEpochPerByte * nProofEpochs * rawSize;
```
ZenGround0, line 51: nit: since per is denoting a rate to be multiplied over a number of things in the `rewardPerEpochPerByte` it would make more sense to drop `Per` and have `periodReward` or `estimatedReward` or just `reward`

src/Fees.sol
```diff
@@ -1,58 +1,69 @@
...
+        } else if (estimatedGasFee >= gasLimitLeft) {
+            return gasLimitRight - estimatedGasFee; // Partial discount on proof fee
         } else {
-            gasPrice = ONE_NANO_FIL;
+            return (rewardPerPeriod * PROOF_FEE_PERCENTAGE) / 100;
         }
-        uint256 numerator = PROOF_GAS_FLOOR * challengeCount * gasPrice;
-        uint256 denominator = ONE_PERCENT_DENOM;
-        return numerator / denominator;
     }
```
ZenGround0, line 65: Since the last iteration this has become so much cleaner thanks everybody for simplifying.

src/Fees.sol
ZenGround0: Ah ok, looking at the docs I think it will typically return a negative exponent and an integer containing lots of precision.  So in this case say FIL = 6.886768 then fiprice will be 6886768 and exponent will be -6 to move the decimal to the correct place.  So in this case the precision will be accounted for.  

Is this how you've seen it work @aarshkshah1992 ?

test/Fees.t.sol
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    function testProofFeeWithGasFeeBoundMediumGasFee() public pure {
+        uint64 filUsdPrice = 5;
+        int32 filUsdPriceExpo = 0;
+        uint256 rawSize = 1e18;
+
+        uint256 rewardPerPeriod = computeRewardPerPeriod(filUsdPrice, filUsdPriceExpo, rawSize);
+
+        uint256 gasLimitLeft = (rewardPerPeriod * PDPFees.GAS_LIMIT_LEFT_PERCENTAGE) / 100;
```
ZenGround0, line 61: nit: put these as contract constants so you don't have to recompute

test/Fees.t.sol
```diff
@@ -0,0 +1,146 @@
...
+        uint256 gasLimitRight = (rewardPerPeriod * PDPFees.GAS_LIMIT_RIGHT_PERCENTAGE) / 100;
+
+        uint256 estimatedGasFee = (gasLimitLeft + gasLimitRight) / 2;
```
ZenGround0, line 64: nit: would be nice to add to this case tests at the boundaries, i.e. estimatedGasFee = gasLimitLeft and estimatedGasFee = gasLimitRight - 1

src/Fees.sol
aarshkshah1992: Yes what you are saying is exactly correct. So we need to handle both positive and negative exponents here.

.github/workflows/makefile.yml
```diff
@@ -16,6 +16,16 @@ jobs:
       - name: configure
         run: ./configure
 
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: nightly
```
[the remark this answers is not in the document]
aarshkshah1992: Yes. But it's all right.

src/PDPVerifier.sol
```diff
@@ -28,12 +29,18 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
     uint256 public constant MAX_ENQUEUED_REMOVALS = 2000;
     address public constant RANDOMNESS_PRECOMPILE = 0xfE00000000000000000000000000000000000006;
     uint256 public constant EXTRA_DATA_MAX_SIZE = 2048;
+    IPyth public constant PYTH = IPyth(0xA2aa501b19aff244D90cc15a4Cf739D2725B5729);
```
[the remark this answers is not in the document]
aarshkshah1992: Yes it's network agnostic. See https://docs.pyth.network/price-feeds/contract-addresses/evm

Makefile
```diff
@@ -8,7 +8,13 @@ CHALLENGE_FINALITY ?=
 
 # Default target
 .PHONY: all
-all: build test
+all: install build test
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

src/PDPVerifier.sol
```diff
@@ -584,4 +631,20 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
     function heightFromIndex(uint256 index) internal pure returns (uint256) {
         return BitOps.ctz(index + 1);
     }
+
+    // Add function to get FIL/USD price
+    function getFILUSDPrice() public returns (uint64, int32) {
+        // Get FIL/USD price no older than 1 day
+        PythStructs.Price memory priceData = PYTH.getPriceNoOlderThan(
+            FIL_USD_PRICE_FEED_ID,
+            86400
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

src/PDPVerifier.sol
```diff
@@ -387,10 +395,49 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
...
+            payable(msg.sender).transfer(msg.value - proofFee);
+        }
+        emit ProofFeePaid(setId, proofFee);
+    }
+
+    function calculateCallDataSize(Proof[] calldata proofs) internal pure returns (uint256) {
+        uint256 callDataSize = 0;
+        for (uint256 i = 0; i < proofs.length; i++) {
+            // 32 for the leaf + each element in the proof is 32 bytes
+            callDataSize += 32 + (proofs[i].proof.length * 32);
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

src/Fees.sol
```diff
@@ -1,58 +1,69 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.20;
 
+import {BitOps} from "./BitOps.sol";
+import "forge-std/console.sol";
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

test/Fees.t.sol
```diff
@@ -0,0 +1,146 @@
...
+        uint64 filUsdPrice = 5000;
+        int32 filUsdPriceExpo = -3;
+        uint256 rawSize = 1e18;
+        uint256 estimatedGasFee = 1e15;
+
+        uint256 fee = PDPFees.proofFeeWithGasFeeBound(estimatedGasFee, filUsdPrice, filUsdPriceExpo, rawSize, epochs_per_day);
+        assertTrue(fee > 0, "Fee should be positive with negative exponent");
+    }
+
+    function testProofFeeWithGasFeeBoundLargeRawSize() public pure {
```
ZenGround0, line 101: I don't understand the point of this test.

src/Fees.sol
```diff
@@ -1,58 +1,65 @@
...
+    uint256 constant SYBIL_FEE = ONE_FIL / 10;
+
+    // assume 1FIL = $5 for now -> we can change this to use an oracle in PDP V1
+    uint256 constant FIL_USD_PRICE = 5;
+
+    // 2 USD/Tib/month is the current reward earned by Storage Providers
+    uint256 constant MONTHLY_TIB_STORAGE_REWARD_USD = 2;
+
+    uint256 constant DAILY_TIB_STORAGE_REWARD_ATTO_FIL =
+        (MONTHLY_TIB_STORAGE_REWARD_USD * 1e18 * ATTO_FIL) /
```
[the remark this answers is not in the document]
aarshkshah1992: This comment is now outdated.

src/Fees.sol
ZenGround0: I think I was just confusing myself imagining price being between 1 and 10 for some reason.  You can make the price be 133 or 16666 just using the price integer.  I'd be curious to see how much precision the oracle gives and what representation they typically choose.  But I'm not concerned about this anymore.

test/PDPVerifier.t.sol
```diff
@@ -593,35 +609,45 @@ contract PDPVerifierProofTest is Test, ProofBuilderHelper {
...
 
-        // Calculate the correct proof fee
-        uint256 correctProofFee = PDPFees.proofFee(challengeCount);
+        // Mock block.number to 2880
+        vm.roll(2880);
 
         // Test 1: Sending less than the required fee
-        uint256 insufficientFee = correctProofFee - 1;
+
+        uint256 correctFee = 388051072754688;
```
ZenGround0, line 644: Thanks, overall better to have this than not I think.  Please leave a comment stating why we need to to it this way.  If there is any more development on proving code we will run into failures every time here so at least we should make it easy to fix them.  Probably a good idea to leave instructions here for that too.

test/PDPVerifier.t.sol
```diff
@@ -693,7 +732,21 @@ contract PDPVerifierProofTest is Test, ProofBuilderHelper {
 
         // Submit proof successfully, advancing the proof set to a new challenge epoch.
         vm.mockCall(pdpVerifier.RANDOMNESS_PRECOMPILE(), abi.encode(challengeEpoch), abi.encode(challengeEpoch));
-        pdpVerifier.provePossession{value: PDPFees.proofFee(proofs.length)}(setId, proofs);
+        // Mock Pyth oracle call to return $5 USD/FIL
+        bytes memory pythCallData = abi.encodeWithSelector(
```
ZenGround0, line 736: These code blocks are repeated many times.  I think you could encapsulate call data generation into a helper function of price and expo.

ZenGround0 (approved):
While I've left a bunch of comments I don't think anything is strictly blocking.  The most important thing is initializing the `proofSetLastProvenEpoch` state variable.

src/PDPVerifier.sol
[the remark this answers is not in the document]
aarshkshah1992: @ZenGround0 

I have implemented the following changes

1) When an new proof set is created -> `proofSetLastProvenEpoch ` is set to ` NO_PROVEN_EPOCH` i.e. 0 (this is the default I know but just making it explicit)

2) When a proof set is deleted -> `proofSetLastProvenEpoch` is set to `NO_PROVEN_EPOCH`

3)When an existing proofset becomes empty after all it's roots are removed -> `proofSetLastProvenEpoch` is set to `NO_PROVEN_EPOCH`

4) When the first root is added for a `proofset` -> set `proofSetLastProvenEpoch` to current block number as this option sounds the most reasonable to me.

src/Fees.sol
aarshkshah1992: Done.

src/Fees.sol
aarshkshah1992: @ZenGround0 I think the exponent is more useful for prices < 1 USD 😅

src/Fees.sol
ZenGround0: The original comment is now out of date given that we are measuring proving time between calls successive proof calls..

src/PDPVerifier.sol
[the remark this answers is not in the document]
ZenGround0: Thanks that sounds perfect, I'll take another look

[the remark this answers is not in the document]
ZenGround0: One final request here is to add a test ensuring that this variable is set correctly when the last root of a proofset is removed

test/PDPVerifier.t.sol
aarshkshah1992: Done.

test/PDPVerifier.t.sol
aarshkshah1992: Done.

test/Fees.t.sol
aarshkshah1992: Done.

test/Fees.t.sol
aarshkshah1992: Not sure what you mean. These need to be computed on the fly as the `rewardPerPeriod` is not a constant.

test/Fees.t.sol
aarshkshah1992: Just a test with the biggest data size (1e30 bytes) among all the tests. Dosent hurt.

aarshkshah1992:
@ZenGround0 Have addressed all your comments. Please can you take a look ?

src/PDPVerifier.sol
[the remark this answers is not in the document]
aarshkshah1992: @ZenGround0 Added a test for this. See `testProofSetLastProvenEpochOnRootRemoval`

aarshkshah1992:
@ZenGround0 Have addressed all your review comments. Please can you give it one more look ?

aarshkshah1992:
@ZenGround0 Merging as discussed offline

PR #84, opened 2024-12-05
feat: SimplePDPService: improve the fault record event

aarshkshah1992:
Add the `proofsetID` and the actual faulted deadlines to the `FaultRecordEvent`.


src/SimplePDPService.sol
```diff
@@ -211,12 +211,16 @@ contract SimplePDPService is PDPListener, PDPRecordKeeper, Initializable, UUPSUp
...
         uint256 faultPeriods = periodsSkipped;
-        if (!provenThisPeriod[proofSetId]) { 
-            // include previous unproven period 
+        if (!provenThisPeriod[proofSetId]) {
+            // include previous unproven period
             faultPeriods += 1;
         }
         if (faultPeriods > 0) {
-            emit FaultRecord(faultPeriods);
+            uint256[] memory faultedPeriodsList = new uint256[](faultPeriods);
```
ZenGround0, line 219: I don't think this is good.  This event should be constant sized.  It is trivial to reconstruct the faultedPeriodsList offchain from constant size.  And then there's no issues with the event growing very big when recovering from long standing faults.  We should do (provingDeadlines[proofSetId], faultPeriods) or (nextDeadline, faultPeriods)

test/SimplePDPService.t.sol
```diff
@@ -163,25 +163,29 @@ contract SimplePDPServiceFaultsTest is Test {
 
     function testFaultWithinOpenPeriod() public {
         pdpService.rootsAdded(proofSetId, 0, new PDPVerifier.RootData[](0), empty);
-
```
ZenGround0, line 166: Where are these noisy edits coming from?  If we want to include them we should be running a linting check on CI and fix all of them at once.

aarshkshah1992:
Closing in favour of https://github.com/FilOzone/pdp/pull/87

PR #85, opened 2024-12-11
Some security relevant comments

ZenGround0:
@irenegia @lucaniz I think there were some more comments of this sort you wanted to include, could you refresh my memory ?

### DIFF
--- a/src/PDPVerifier.sol
+++ b/src/PDPVerifier.sol
@@ -74,8 +74,9 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
     // randomness sampling for challenge generation.
     //
     // The purpose of this delay is to prevent SPs from biasing randomness by running forking attacks.
-    // This is actually not possible with the challenge sampling method written here. Qe sample from DRAND
-    // and forking attacks are unrelated to biasability, hence challengeFinality = 1 is a safe value.
+    // Given a small enough challengeFinality an SP can run several trials of challenge sampling and 
+    // fork around samples that don't suit them, grinding the challenge randomness.
+    // For the filecoin L1, a safe value is 150.
     //
     // We keep this around for future portability to a variety of environments with different assumptions
     // behind their challenge randomness sampling methods.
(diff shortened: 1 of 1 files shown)


aarshkshah1992:
@ZenGround0 But where are we changing the value to 150 ? Shouldn't this PR also change the deployment scripts to change it to 150 instead of 1 ?

src/PDPVerifier.sol
```diff
@@ -74,8 +74,9 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
     // randomness sampling for challenge generation.
     //
     // The purpose of this delay is to prevent SPs from biasing randomness by running forking attacks.
-    // This is actually not possible with the challenge sampling method written here. Qe sample from DRAND
-    // and forking attacks are unrelated to biasability, hence challengeFinality = 1 is a safe value.
+    // Given a small enough challengeFinality an SP can run several trials of challenge sampling and 
+    // fork around samples that don't suit them, grinding the challenge randomness.
+    // For the filecoin L1, a safe value is 150.
```
lucaniz, line 79: // We are setting here this value at 150, as in the Filecoin L1 (epochs between PreCommit and ProveCommit for interactive PoRep Sectors)

src/PDPVerifier.sol
```diff
@@ -383,6 +384,13 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         uint256 sumTreeTop = 256 - BitOps.clz(nextRootId[setId]);
         for (uint64 i = 0; i < proofs.length; i++) {
             // Hash (SHA3) the seed,  proof set id, and proof index to create challenge.
+            // Note -- there is a slight deviation here from the uniform distribution.
+            // Some leaves are challenged with probability p and some have probability p + deviation. 
+            // This deviation is bounded by leafCount / 2^256 given a 256 bit hash
+            // Assuming a 1000EiB = 1 ZiB network size ~ 2^70 bytes of data or 2^65 leaves
+            // This deviation is bounded by 2^65 / 2^256 = 2^-191 which is negligible.            
+            //   If modifying this code to use a hash function with smaller output size
```
lucaniz, line 392: i'd add a comment on which is the "standard solution" (i.e. rejection sampling). Best would be to have a quick note on why we are not implementing it

ZenGround0:
> But where are we changing the value to 150 ? Shouldn't this PR also change the deployment scripts to change it to 150 instead of 1 ?

Yes good point

aarshkshah1992 (approved):

PR #86, opened 2024-12-12
Remove special casing of first add

ZenGround0:
Closes #

- [x] Remove special casing of first add
- [x] Handle first add in next proving period 
- [x] Update PDP service to support this behavior
- [x] Update deadline calculator functions to return useful values when state not yet initialized
- [x] Support graceful stopping of PDP Service accounting when all roots are removed

### DIFF
--- a/src/IPDPProvingSchedule.sol
+++ b/src/IPDPProvingSchedule.sol
@@ -12,6 +12,11 @@ interface IPDPProvingSchedule {
     /// @return Challenge window size in epochs
     function challengeWindow() external pure returns (uint256);
 
+    /// @notice Value for initializing the challenge window start for a given proof set assuming proving period starts now
+    // @return Initial challenge window start in epochs
+    function initChallengeWindowStart(uint256 setId) external pure returns (uint256);
+
+
     /// @notice Calculates the start of the next challenge window for a given proof set
     /// @param setId The ID of the proof set
     /// @return The block number when the next challenge window starts
--- a/src/PDPVerifier.sol
+++ b/src/PDPVerifier.sol
@@ -477,6 +477,11 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
     function nextProvingPeriod(uint256 setId, uint256 challengeEpoch, bytes calldata extraData) public {
         require(extraData.length <= EXTRA_DATA_MAX_SIZE, "Extra data too large");
         require(msg.sender == proofSetOwner[setId], "only the owner can move to next proving period");
+        
+        if (proofSetLastProvenEpoch[setId] == NO_PROVEN_EPOCH) {
+            proofSetLastProvenEpoch[setId] = block.number;
+        }
+
         // Take removed roots out of proving set
         uint256[] storage removals = scheduledRemovals[setId];
         uint256[] memory removalsToProcess = new uint256[](removals.length);
--- a/src/PDPVerifier.sol
+++ b/src/PDPVerifier.sol
@@ -34,6 +34,7 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
 
     // FIL/USD price feed query ID on the Pyth network
     bytes32 public constant FIL_USD_PRICE_FEED_ID = 0x150ac9b959aee0051e4091f0ef5216d941f590e1c5e7f91cf7635b5c11628c0e;
+    uint256 public constant NO_CHALLENGE_SCHEDULED = 0;
 
     // Events
     event ProofSetCreated(uint256 indexed setId);
@@ -264,7 +265,7 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
 
         uint256 setId = nextProofSetId++;
         proofSetLeafCount[setId] = 0;
-        nextChallengeEpoch[setId] = 0;  // Re-initialized when the first root is added.
+        nextChallengeEpoch[setId] = NO_CHALLENGE_SCHEDULED;  // Initialized on first call to NextProvingPeriod
         proofSetOwner[setId] = msg.sender;
         proofSetListener[setId] = listenerAddr;
         proofSetLastProvenEpoch[setId] = NO_PROVEN_EPOCH;
@@ -310,17 +311,10 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         require(proofSetLive(setId), "Proof set not live");
         require(rootData.length > 0, "Must add at least one root");
         require(proofSetOwner[setId] == msg.sender, "Only the owner can add roots");
-        bool needsChallengeEpoch = nextChallengeEpoch[setId] == 0;
         uint256 firstAdded = nextRootId[setId];
 
         for (uint256 i = 0; i < rootData.length; i++) {
             addOneRoot(setId, i, rootData[i].root, rootData[i].rawSize);
-        }
-        // Initialise the first challenge epoch and challengeable leaf range when the first data is added.
-        if (needsChallengeEpoch) {
-            nextChallengeEpoch[setId] = block.number + challengeFinality;
-            challengeRange[setId] = proofSetLeafCount[setId];
-            proofSetLastProvenEpoch[setId] = block.number;
         }
 
         address listenerAddr = proofSetListener[setId];
@@ -388,6 +382,8 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         uint256 challengeEpoch = nextChallengeEpoch[setId];
         require(block.number >= challengeEpoch, "premature proof");
         require(proofs.length > 0, "empty proof");
+        require(challengeEpoch != NO_CHALLENGE_SCHEDULED, "no challenge scheduled");
+        require(proofSetOwner[setId] == msg.sender, "only the owner can prove possession");
 
         uint256 seed = drawChallengeSeed(setId);
         uint256 leafCount = challengeRange[setId];
@@ -493,21 +489,21 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         removeRoots(setId, removalsToProcess);
         // Bring added roots into proving set
         challengeRange[setId] = proofSetLeafCount[setId];
-        if (challengeEpoch - block.number < challengeFinality) {
+        if (challengeEpoch < block.number + challengeFinality) {
             revert("challenge epoch must be at least challengeFinality epochs in the future");
         }
         nextChallengeEpoch[setId] = challengeEpoch;
 
         // Clear next challenge epoch if the set is now empty.
-        // It will be re-set when new data is added.
+        // It will be re-set after new data is added and nextProvingPeriod is called.
         if (proofSetLeafCount[setId] == 0) {
             proofSetLastProvenEpoch[setId] = NO_PROVEN_EPOCH;
-            nextChallengeEpoch[setId] = 0;
+            nextChallengeEpoch[setId] = NO_CHALLENGE_SCHEDULED;
         }
 
         address listenerAddr = proofSetListener[setId];
         if (listenerAddr != address(0)) {
-            PDPListener(listenerAddr).nextProvingPeriod(setId, challengeEpoch,proofSetLeafCount[setId], extraData);
+            PDPListener(listenerAddr).nextProvingPeriod(setId, nextChallengeEpoch[setId], proofSetLeafCount[setId], extraData);
         }
     }
 
--- a/src/PDPVerifier.sol
+++ b/src/PDPVerifier.sol
@@ -42,6 +42,7 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
     event RootsAdded(uint256 indexed firstAdded);
     event RootsRemoved(uint256[] indexed rootIds);
     event ProofFeePaid(uint256 indexed setId, uint256 fee, uint64 price, int32 expo);
+    event ProofSetEmpty(uint256 indexed setId);
 
     // Types
     // State fields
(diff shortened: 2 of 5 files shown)


src/PDPVerifier.sol
```diff
@@ -481,6 +477,11 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
     function nextProvingPeriod(uint256 setId, uint256 challengeEpoch, bytes calldata extraData) public {
         require(extraData.length <= EXTRA_DATA_MAX_SIZE, "Extra data too large");
         require(msg.sender == proofSetOwner[setId], "only the owner can move to next proving period");
+        
+        if (proofSetLastProvenEpoch[setId] == NO_PROVEN_EPOCH) {
+            proofSetLastProvenEpoch[setId] = block.number;
```
aarshkshah1992, line 482: @ZenGround0 I am not sure this is correct. This should only be updated in `provePossesion` after it is successful.
What's the rationale here ?

src/PDPVerifier.sol
```diff
@@ -388,6 +382,8 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         uint256 challengeEpoch = nextChallengeEpoch[setId];
         require(block.number >= challengeEpoch, "premature proof");
         require(proofs.length > 0, "empty proof");
```
aarshkshah1992, line 384: @ZenGround0 We should also revert `provePossession ` if `challengeRange` is 0.

src/PDPVerifier.sol
```diff
@@ -310,18 +311,11 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         require(proofSetLive(setId), "Proof set not live");
         require(rootData.length > 0, "Must add at least one root");
         require(proofSetOwner[setId] == msg.sender, "Only the owner can add roots");
-        bool needsChallengeEpoch = nextChallengeEpoch[setId] == 0;
         uint256 firstAdded = nextRootId[setId];
 
         for (uint256 i = 0; i < rootData.length; i++) {
             addOneRoot(setId, i, rootData[i].root, rootData[i].rawSize);
         }
-        // Initialise the first challenge epoch and challengeable leaf range when the first data is added.
```
aarshkshah1992, line 319: Would you be good to document on `addRoots` that a prover HAS to call `nextProvingPeriod` before these trees are eligible for proving.

src/PDPVerifier.sol
aarshkshah1992: Also `provePossession` should also assert that the current epoch > challengeEpoch

src/PDPVerifier.sol
```diff
@@ -481,6 +477,11 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
     function nextProvingPeriod(uint256 setId, uint256 challengeEpoch, bytes calldata extraData) public {
         require(extraData.length <= EXTRA_DATA_MAX_SIZE, "Extra data too large");
         require(msg.sender == proofSetOwner[setId], "only the owner can move to next proving period");
```
aarshkshah1992, line 479: @ZenGround0 We should revert this function if prover calls it on an empty proofset.  Prover MUST make atleast one call to `addRoot` before calling this to get set up their first deadline to start proving.

src/PDPVerifier.sol
aarshkshah1992: Why not just initialise this the first time a root is added to a proof set and then only update it on successful calls to `provepossesion` ?

src/SimplePDPService.sol
```diff
@@ -112,9 +115,10 @@ contract SimplePDPService is PDPListener, PDPRecordKeeper, Initializable, UUPSUp
     }
 
     // The start of the challenge window for the current proving period
+    // If proof set state not yet initialized assume proving period starts now
     function thisChallengeWindowStart(uint256 setId) public view returns (uint256) {
-        if (provingDeadlines[setId] == 0) {
-            revert("Proving not yet started");
+        if (provingDeadlines[setId] == NO_PROVING_DEADLINE) {
+            return block.number + getMaxProvingPeriod() - challengeWindow();
```
aarshkshah1992, line 121: What's the rationale for not reverting here anymore ?

src/SimplePDPService.sol
```diff
@@ -112,9 +115,10 @@ contract SimplePDPService is PDPListener, PDPRecordKeeper, Initializable, UUPSUp
     }
 
     // The start of the challenge window for the current proving period
+    // If proof set state not yet initialized assume proving period starts now
```
aarshkshah1992, line 118: Why ? Can you explain what goes wrong if we don't do this ?

src/SimplePDPService.sol
ZenGround0: The idea is to make this as useful as possible in the error case.  In curio code I keep getting thwarted by this error when what I really want is for it to return to me the value I should call to nextProvingPeriod.  You could argue maybe this should belong in a second function instead and I'm still mulling that over.

src/SimplePDPService.sol
ZenGround0: I can't get the info I need as SP software consuming this to start up my proofset.  Would you prefer putting this in a separate function?

src/SimplePDPService.sol
```diff
@@ -177,6 +179,9 @@ contract SimplePDPService is PDPListener, PDPRecordKeeper, Initializable, UUPSUp
         if (challengeCount < getChallengesPerProof()) {
             revert("Invalid challenge count < 5");
         }
+        if (provingDeadlines[proofSetId] == NO_PROVING_DEADLINE) {
+            revert("Proving not yet started");
```
aarshkshah1992, line 183: @ZenGround0 After thinking through it, I also think it's a good idea to revert this function if a proof does not fall in the challenge window i.e. in [0,`challengeWindow()`] epochs before the proving deadline.

src/PDPVerifier.sol
```diff
@@ -493,21 +494,21 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
...
         }
         nextChallengeEpoch[setId] = challengeEpoch;
 
         // Clear next challenge epoch if the set is now empty.
-        // It will be re-set when new data is added.
+        // It will be re-set after new data is added and nextProvingPeriod is called.
         if (proofSetLeafCount[setId] == 0) {
             proofSetLastProvenEpoch[setId] = NO_PROVEN_EPOCH;
-            nextChallengeEpoch[setId] = 0;
+            nextChallengeEpoch[setId] = NO_CHALLENGE_SCHEDULED;
```
aarshkshah1992, line 506: I think we should revert if the prover passes in a non-zero `challengeEpoch` for a proofset that becomes empty after processing all their removals. It's not a valid path for the prover to take.

src/SimplePDPService.sol
```diff
@@ -190,24 +195,38 @@ contract SimplePDPService is PDPListener, PDPRecordKeeper, Initializable, UUPSUp
...
             periodsSkipped = (block.number - (provingDeadlines[proofSetId] + 1)) / getMaxProvingPeriod();
         }
-        // ensure next challenge epoch falls within the next challenge window.
-        // The next challenge window immediately precedes the next deadline
+
         uint256 nextDeadline = provingDeadlines[proofSetId] + getMaxProvingPeriod()*(periodsSkipped+1);
-        if (challengeEpoch < nextDeadline - challengeWindow() || challengeEpoch > nextDeadline) {
+        // the proofset has become empty and provingDeadline is set inactive 
+        if (challengeEpoch == NO_CHALLENGE_SCHEDULED) {
+            nextDeadline = NO_PROVING_DEADLINE;
```
aarshkshah1992, line 228: Do the calculation on line 225 above only if this branch is not hit

src/PDPVerifier.sol
aarshkshah1992: It does not make sense to start "proving" an "empty proofset".

aarshkshah1992 (commented):
@ZenGround0 Left first round of comments. Will do one more round once you address these.

src/PDPVerifier.sol
ZenGround0: This check would be correct but I don't think I like this because this check is covered by requiring that challengeEpoch != NO_CHALLENGE_SCHEDULED

The only way to schedule a challenge is to call nextProvingPeriod which sets challengeRange equal to leafCount.  If the leafCount is at zero nextProvingPeriod will revert challengeEpoch back to NO_CHALLENGE_SCHEDULED.

Can you see any edge case around this?  If there is I think I'll want to do some more thinking about the design we've got.

src/PDPVerifier.sol
ZenGround0: I like this

src/SimplePDPService.sol
ZenGround0: I think I agree.  It is also redundant and I worry it will confuse people down the line but I can add a longer comment above the check.

src/PDPVerifier.sol
ZenGround0: Yeah I think this is a more consistent style

ZenGround0: But thinking more about it I prefer it the way it is.  By the time we get to the code to figure out if we should revert we've already done most of the work.  And the code has enough info to do the right thing in this case so it should do the right thing.  One thing that would be helpful though is being very explicit about this to the caller.  I think emitting an event here would be nice.

src/SimplePDPService.sol
ZenGround0: I'm going to make another function for this behavior.

src/SimplePDPService.sol
ZenGround0: Moving this functionality to `initChallengeWindowStart`

src/SimplePDPService.sol
```diff
@@ -190,24 +195,38 @@ contract SimplePDPService is PDPListener, PDPRecordKeeper, Initializable, UUPSUp
     // nextProvingPeriod checks for unsubmitted proof and emits a fault if so
     function nextProvingPeriod(uint256 proofSetId, uint256 challengeEpoch, uint256 leafCount, bytes calldata) external onlyPDPVerifier {
         receiveProofSetEvent(proofSetId, OperationType.NEXT_PROVING_PERIOD, abi.encode(challengeEpoch, leafCount));
-        // Noop when proving period not yet open
+
+        // initialize state for new proofset
+        if (provingDeadlines[proofSetId] == NO_PROVING_DEADLINE) {
```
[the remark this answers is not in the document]
ZenGround0: Can you explain what we gain from this?  Since we're checking that the epoch is within the challenge window that will already cover this check right?

src/SimplePDPService.sol
ZenGround0: Do you like 
```solidity
        uint256 nextDeadline;
        // the proofset has become empty and provingDeadline is set inactive 
        if (challengeEpoch == NO_CHALLENGE_SCHEDULED) {
            nextDeadline = NO_PROVING_DEADLINE;
        } else {
            nextDeadline = provingDeadlines[proofSetId] + getMaxProvingPeriod()*(periodsSkipped+1);
            if (challengeEpoch < nextDeadline - challengeWindow() || challengeEpoch > nextDeadline) {
                revert("Next challenge epoch must fall within the next challenge window");
            }
        }
```
better?  I could see this being a bit more clear but with the extra nested if I'm not sure.

src/PDPVerifier.sol
aarshkshah1992: Okay makes sense. 

Also, I was worried that this means that the prover now has to pay low proof fees even if they fault certain periods but realise now that we only update `proofSetLastProvenEpoch` here if it is 0. This means that this only applies to the transition of going from an proofset not being proven at all to it being proven and that is fine.

src/PDPVerifier.sol
aarshkshah1992: Yes but what's the harm in putting this in place to make it explicit ? I can't see an edge case around this but want to make it more explicit here.

aarshkshah1992: Your call, no strong opinions.

src/SimplePDPService.sol
aarshkshah1992: This is much better

aarshkshah1992: I mean the new code you have proposed here is much better.

src/SimplePDPService.sol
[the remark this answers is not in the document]
aarshkshah1992: okay yeah fair enough. Didnt realise this.

src/SimplePDPService.sol
ZenGround0: Done

src/SimplePDPService.sol
ZenGround0: Done

src/SimplePDPService.sol
ZenGround0: Done

src/PDPVerifier.sol
ZenGround0: Event now in place

src/IPDPProvingSchedule.sol
```diff
@@ -12,6 +12,11 @@ interface IPDPProvingSchedule {
     /// @return Challenge window size in epochs
     function challengeWindow() external pure returns (uint256);
 
+    /// @notice Value for initializing the challenge window start for a given proof set assuming proving period starts now
+    // @return Initial challenge window start in epochs
+    function initChallengeWindowStart(uint256 setId) external pure returns (uint256);
```
aarshkshah1992, line 17: nice change

src/PDPVerifier.sol
```diff
@@ -388,6 +384,8 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         uint256 challengeEpoch = nextChallengeEpoch[setId];
         require(block.number >= challengeEpoch, "premature proof");
         require(proofs.length > 0, "empty proof");
+        require(challengeEpoch != NO_CHALLENGE_SCHEDULED, "no challenge scheduled");
+        require(proofSetOwner[setId] == msg.sender, "only the owner can prove possession");
```
aarshkshah1992, line 388: @ZenGround0 This check already exists above and is a duplicate.

aarshkshah1992 (approved):
@ZenGround0 Lgtm. Just need to fix the duplicate check for owner in `provePossesion` in the `PDPVerifier` contract.

src/PDPVerifier.sol
ZenGround0: Thanks

PR #87, opened 2024-12-13
fix: emit fault record event with more details

aarshkshah1992:
Replaces https://github.com/FilOzone/pdp/pull/84/files.

### DIFF
--- a/src/SimplePDPService.sol
+++ b/src/SimplePDPService.sol
@@ -17,7 +17,7 @@ contract PDPRecordKeeper {
         REMOVE_SCHEDULED,
         PROVE_POSSESSION,
         NEXT_PROVING_PERIOD
-    }    
+    }
 
     // Struct to store event details
     struct EventRecord {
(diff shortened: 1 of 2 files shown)


src/SimplePDPService.sol
```diff
@@ -171,16 +171,16 @@ contract SimplePDPService is PDPListener, PDPRecordKeeper, Initializable, UUPSUp
     // it also checks that proofs are not late and emits a fault record if so
     function possessionProven(uint256 proofSetId, uint256 challengedLeafCount, uint256 seed, uint256 challengeCount) external onlyPDPVerifier {
         receiveProofSetEvent(proofSetId, OperationType.PROVE_POSSESSION, abi.encode(challengedLeafCount, seed, challengeCount));
-        if (provenThisPeriod[proofSetId]) { 
-            revert("Only one proof of possession allowed per proving period. Open a new proving period.");
```
aarshkshah1992, line 175: Hmm these extra spacing changes are frustrating. Sorry. Will dive deep into my editor settings today to see why this is happening.

ZenGround0: At this point I'm getting used to it so don't worry too much about it

ZenGround0 (approved):

src/SimplePDPService.sol
aarshkshah1992: @ZenGround0 Found the root cause in my editor and fixed it. I'll add a formatter to our CI so we can stop worrying about it.

PR #92, opened 2025-01-24
Fix PDP events UX

### LINKED ISSUE #79: Allow Spark to observe added/removed roots
Hello from the Spark team :wave:

I reviewed the [docs](https://filoznotebook.notion.site/Proof-of-Data-Possession-4960f7827f074748a7e0227381455ad2) and the [PDPVerifier contract](https://github.com/FilOzone/pdp/blob/19993af01548f90a8ee0d8822f6ce885dd1cc16d/src/PDPVerifier.sol) to understand how can Spark maintain a list of Root CIDs that are covered by PDP checks.

I found the following four events - I think they cover all state changes relevant to us 👍🏻 

https://github.com/FilOzone/pdp/blob/19993af01548f90a8ee0d8822f6ce885dd1cc16d/src/PDPVerifier.sol#L31-L34

I can imagine how a PDP observer can use `ProofSetCreated` and `ProofSetDeleted` events to maintain a view of all active proof sets 👍🏻 

However, I don't understand how we can maintain a view of all active roots.

**Question 1: observing when roots are added to a root-set**

https://github.com/FilOzone/pdp/blob/14fd1c9225471aab4a517abdaf4791c8d9963fd4/contracts/src/PDPVerifier.sol#L280-L282

Let's say somebody calls `addRoots()` with the following arguments:
```
setId=1
rootData=[ {cid: 'baf1', rawSize: 64}, {cid: 'bafy2', rawSize: 256 } ]
```

The event `RootsAdded` provides only `firstAdded` argument, which seems to be the `rootId` index into the three per-proof-set mappings:
https://github.com/FilOzone/pdp/blob/19993af01548f90a8ee0d8822f6ce885dd1cc16d/src/PDPVerifier.sol#L323-L325

Using my example above, how can I reconstruct `setId=1` and `rootData=[ {cid: 'baf1', rawSize: 64}, {cid: 'bafy2', rawSize: 256 } ]` from the `firstAdded` field reported by the `RootsAdded` event?

I _think_ you need to add at least `setId` field to the event payload.

If we can assume that `addRoots()` will map individual roots to root ids that are in sequence (i.e. `firstRoot`, `firstRoot+1`, ... `firstRoot+N-1`), then the combination of `setId` and `firstRoot` allows the observer to iterate over all roots. The remaining question is how to know when to stop the iteration (what was the length of the `rootData` array).

If my reasoning is correct, then we need to change the event signature to the following one:

```solidity
event RootsAdded(uint256 indexed setId, uint256 indexed firstAdded, uint256 addedCount);
```

While thinking about this more:

(1)
I don't understand why is `firstAdded` indexed - how do you envision the client queries that will use this index? It makes sense to me to look up events linked to a specific root set (via `setId`). I struggle to find a use case for looking up the events using the firstAdded value. (If I wanted to look up the event by root id, then how would I go from a particular root id to the "firstAdded" value emitted when the root was aded?)

Proposal:

```solidity
event RootsAdded(uint256 indexed setId, uint256 firstAdded, uint256 addedCount);
```

(2)
I find the name "firstAdded" very confusing, it does not tell me how to interpret the value. Something like `idOfTheFirstRootAdded` is easier to understand, but also rather verbose. 

Here is a suggestion that works well for me:

```solidity
event RootsAdded(uint256 indexed setId, uint256 firstRootId, uint256 lastRootId);
```

Then I can run the following code to reconstruct the root CIDs & sizes that were added to the proof set:

```js
for (let rootId = firstRootId; rootId <= lastRootId; rootId++) {
  const cid = rootCids[setId][rootId]
  const size = rootLeafCounts[setId][rootId] * LEAF_SIZE
  // add {rootId, cid, size} to our view of roots
}
```

Alternatively, can we model the `RootsAdded` event after `RootsRemoved` event that provides an array of root ids? I think that's the most elegant solution. (But I understand it could be more expensive on the gas fees.)

```solidity
event RootsAdded(uint256 indexed setId, uint256[] rootIds);
```

**Question 2: observing when roots are removed from a root-set**

https://github.com/FilOzone/pdp/blob/14fd1c9225471aab4a517abdaf4791c8d9963fd4/contracts/src/PDPVerifier.sol#L329-L331

This is similar to the previous question.

First, the event `RootsRemoved` does not seem to be emitted at all. Huh?? There is only one source code line containing the string `RootsRemoved`:
https://github.com/search?q=repo%3AFilOzone%2Fpdp%20RootsRemoved&type=code

Let's say somebody calls `scheduleRemovals()` with the following arguments:

```
setId=1
rootIds = [10, 20]
```

How can a party observing PDP state learn:
- Which setId+rootId pairs were removed?
- When exactly are these roots considered as no longer covered by PDP proof? Is it when the removal is scheduled or when the roots are actually removed at the start of the next proving period?

aarshkshah1992:
Closes #79 .
Add more information to events and make them easy to consume.

### DIFF
--- a/src/PDPVerifier.sol
+++ b/src/PDPVerifier.sol
@@ -47,6 +47,9 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
    
     event ProofFeePaid(uint256 indexed setId, uint256 fee, uint64 price, int32 expo);
 
+    event PossessionProven(uint256 indexed setId, uint256 challengedLeafCount, uint256 seed, uint256 challengeCount);
+    event NextProvingPeriod(uint256 indexed setId, uint256 challengeEpoch, uint256 leafCount);
+
     // Types
     // State fields
      event Debug(string message, uint256 value);
@@ -436,6 +439,7 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
             PDPListener(listenerAddr).possessionProven(setId, proofSetLeafCount[setId], seed, proofs.length);
         }
         proofSetLastProvenEpoch[setId] = block.number;
+        emit PossessionProven(setId, proofSetLeafCount[setId], seed, proofs.length);
     }
 
     function calculateAndBurnProofFee(uint256 setId, uint256 gasUsed) internal {
@@ -537,6 +541,7 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         if (listenerAddr != address(0)) {
             PDPListener(listenerAddr).nextProvingPeriod(setId, nextChallengeEpoch[setId], proofSetLeafCount[setId], extraData);
         }
+        emit NextProvingPeriod(setId, challengeEpoch, proofSetLeafCount[setId]);
     }
 
     // removes roots from a proof set's state.
--- a/src/PDPVerifier.sol
+++ b/src/PDPVerifier.sol
@@ -47,7 +47,8 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
    
     event ProofFeePaid(uint256 indexed setId, uint256 fee, uint64 price, int32 expo);
 
-    event PossessionProven(uint256 indexed setId, uint256 challengedLeafCount, uint256 seed, uint256 challengeCount);
+
+    event PossessionProven(uint256 indexed setId, RootIdAndOffset[] challenges);
     event NextProvingPeriod(uint256 indexed setId, uint256 challengeEpoch, uint256 leafCount);
 
     // Types
@@ -400,6 +401,7 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         require(block.number >= challengeEpoch, "premature proof");
         require(proofs.length > 0, "empty proof");
         require(challengeEpoch != NO_CHALLENGE_SCHEDULED, "no challenge scheduled");
+        RootIdAndOffset[] memory challenges = new RootIdAndOffset[](proofs.length);
         
         uint256 seed = drawChallengeSeed(setId);
         uint256 leafCount = challengeRange[setId];
@@ -420,12 +422,13 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
             uint256 challengeIdx = uint256(keccak256(payload)) % leafCount;
 
             // Find the root that has this leaf, and the offset of the leaf within that root.
-            RootIdAndOffset memory root = findOneRootId(setId, challengeIdx, sumTreeTop);
-            bytes32 rootHash = Cids.digestFromCid(getRootCid(setId, root.rootId));
-            bool ok = MerkleVerify.verify(proofs[i].proof, rootHash, proofs[i].leaf, root.offset);
+            challenges[i] = findOneRootId(setId, challengeIdx, sumTreeTop);
+            bytes32 rootHash = Cids.digestFromCid(getRootCid(setId, challenges[i].rootId));
+            bool ok = MerkleVerify.verify(proofs[i].proof, rootHash, proofs[i].leaf, challenges[i].offset);
             require(ok, "proof did not verify");
         }
 
+     
         // Note: We don't want to include gas spent on the listener call in the fee calculation
         // to only account for proof verification fees and avoid gamability by getting the listener
         // to do extraneous work just to inflate the gas fee.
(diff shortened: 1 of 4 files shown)


ZenGround0 (commented):
@bajtos any issues with this from your perspective

bajtos:
Awesome, thank you for addressing my feedback! I'll review the changes next week.

src/PDPVerifier.sol
```diff
@@ -37,13 +37,19 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
...
-    event RootsRemoved(uint256[] indexed rootIds);
-    event ProofFeePaid(uint256 indexed setId, uint256 fee, uint64 price, int32 expo);
     event ProofSetEmpty(uint256 indexed setId);
 
+    event RootsAdded(uint256 indexed setId, uint256[] rootIds);
+    event RootsRemoved(uint256 indexed setId, uint256[] rootIds);
+   
+    event ProofFeePaid(uint256 indexed setId, uint256 fee, uint64 price, int32 expo);
+
+    event PossessionProven(uint256 indexed setId, uint256 challengedLeafCount, uint256 seed, uint256 challengeCount);
```
ZenGround0, line 50: I think what we really want is actually a list of (root indexes, root offsets) that are challenged: 
```solidity
event PossessionProven(uint256 indexed setId, RootIdAndOffset[] challenges)
```

src/PDPVerifier.sol
```diff
@@ -321,11 +329,16 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
             addOneRoot(setId, i, rootData[i].root, rootData[i].rawSize);
         }
 
+        uint256[] memory rootIds = new uint256[](rootData.length);
+        for (uint256 i = 0; i < rootData.length; i++) {
+            rootIds[i] = firstAdded + i;
```
ZenGround0, line 334: Any good reason not to do this inside the other loop?

ZenGround0 (commented):
lgtm

src/PDPVerifier.sol
aarshkshah1992: Fixed.

src/PDPVerifier.sol
aarshkshah1992: moved this

src/PDPVerifier.sol
```diff
@@ -316,16 +325,19 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
...
         require(proofSetOwner[setId] == msg.sender, "Only the owner can add roots");
         uint256 firstAdded = nextRootId[setId];
+        uint256[] memory rootIds = new uint256[](rootData.length);
+
 
         for (uint256 i = 0; i < rootData.length; i++) {
             addOneRoot(setId, i, rootData[i].root, rootData[i].rawSize);
+            rootIds[i] = firstAdded + i;
         }
+        emit RootsAdded(setId, rootIds);
```
bajtos, line 335: Comparing this event to the PDPListener event emitted below:

```solidity
PDPListener(listenerAddr).rootsAdded(setId, firstAdded, rootData, extraData);
```

I see PDPListener is also receiving `extraData`.

What is the purpose of that extra data? Would it make sense to include it in the `RootsAdded` event too?

```solidity
emit RootsAdded(setId, rootIds, extraData);
```

I don't see any obvious use case for that extra data in Spark; this may be a YAGNI thing. I am mentioning it here for your consideration.

ZenGround0 (approved):

src/PDPVerifier.sol
```diff
@@ -525,6 +542,7 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         if (listenerAddr != address(0)) {
             PDPListener(listenerAddr).nextProvingPeriod(setId, nextChallengeEpoch[setId], proofSetLeafCount[setId], extraData);
         }
+        emit NextProvingPeriod(setId, challengeEpoch, proofSetLeafCount[setId]);
```
bajtos, line 545: Similar comment applies here - should we include `extraData` in the event payload?

bajtos (approved):
I reviewed what I wrote in https://github.com/FilOzone/pdp/issues/79 and this pull request addresses all my comments. Thank you! ❤️

src/PDPVerifier.sol
aarshkshah1992: @bajtos That `extraData` is for an end user such as Storacha to get "some" data as a callback to it's listener/service (auth etc.) that it negotiated offchain with the client. I don't think we need it in events for now.

It's also not currently obvious if the services/listeners themselves will need this extra data.

ZenGround0: Its safe to assume that extra data reporting can be handled by the listener.

bajtos: Makes sense, thanks for the explanation 👍🏻

PR #93, opened 2025-01-24
Allow fee safety checks to pass when gas estimating

ZenGround0:
We were failing on estimation as base fee is set to 0


aarshkshah1992:
@ZenGround0 Need to fix CI. otherwise lgtm.

aarshkshah1992 (approved):

PR #126, opened 2025-03-28
fix: avoid refund re-entrancy issues by refunding at the bottom

Stebalien:
This turns these refunds into "tail calls" where we guarantee that we make no changes to state after processing the refund. This prevents any re-entrancy issues in _this_ contract.

In theory, this could trigger a re-entrancy issue in the calling contract, but that's not our problem (it's something we'll just need to document).

### DIFF
--- a/test/PDPVerifier.t.sol
+++ b/test/PDPVerifier.t.sol
@@ -683,6 +683,10 @@ contract PDPVerifierProofTest is Test, ProofBuilderHelper {
         price.price = 1;
         vm.mockCall(address(pdpVerifier.PYTH()), pythCallData, abi.encode(price));
 
+        address sender = makeAddr("sender");
+        vm.deal(sender, 1000 ether);
+        vm.startPrank(sender);
+
         uint leafCount = 10;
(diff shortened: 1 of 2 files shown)


Stebalien:
See each commit for why I made the changes I did.

test/PDPVerifier.t.sol
```diff
@@ -699,20 +703,24 @@ contract PDPVerifierProofTest is Test, ProofBuilderHelper {
         // Mock block.number to 2881
         vm.roll(2881);
 
-        // Test 1: Sending less than the required fee
+        // Determine the correct fee.
```
ZenGround0, line 706: Damn this is real nice

ZenGround0 (approved):
Thanks @Stebalien the stack depth limiting and test fixup are both very appreciated.  Sorry again I missed this.  I woke up this morning assuming no one would come in and do this on a friday afternoon and forgot to check entirely.

test/PDPVerifier.t.sol
aarshkshah1992: Yeah TIL !!

PR #127, opened 2025-03-30
Audit 3.3 and 3.4 -- fix refund issues

ZenGround0:
This addresses the audit review for items 3.3 and 3.4 

3.3 
The audit notes that transfers have issues interacting with smart wallet / more sophisticated sending contracts.  A large reason for this is that transfer puts a very restricted limit on gas usage in the transfer function.   This is a blanket reentrency protection that we don't need.  So this PR forwards full gas limit to the transfer to maximize sending contract compatibility (commit `d6b0a639581839eb2fabdf6b3f368f6a372ac6bd`)

3.4
The audit notes that there is a reentrancy issue since the prove possession transfer is not matching the standard Checks Effects Interactions pattern.  To rememdy this we move the refund call to the end of the `provePossession` call (`70dbb0e839220803fe724ef498cc594256110324`).  We do the same for proofSetCreation (`4b04bc08ceae4fa45775556328eaddf1a27ff6de`).

### DIFF
--- a/src/PDPVerifier.sol
+++ b/src/PDPVerifier.sol
@@ -275,10 +275,7 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         uint256 sybilFee = PDPFees.sybilFee();
         require(msg.value >= sybilFee, "sybil fee not met");
         burnFee(sybilFee);
-        if (msg.value > sybilFee) {
-            // Return the overpayment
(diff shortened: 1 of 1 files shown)


Stebalien:
https://github.com/FilOzone/pdp/pull/126?

ZenGround0:
Oh man Im quite sorry I'll merge that one then take anything remaining and leave it up here.

ZenGround0:
Shutting this down for a simpler 3.3 fix on main

PR #130, opened 2025-03-31
Audit 3.3 Allow nontrivial gas for refund

ZenGround0:
3.3

The audit notes that transfers have issues interacting with smart wallet / more sophisticated sending contracts. A large reason for this is that transfer puts a very restricted limit on gas usage in the transfer function. This is a blanket reentrency protection that we don't need. So this PR forwards full gas limit to the transfer to maximize sending contract compatibility.

### DIFF
--- a/src/PDPVerifier.sol
+++ b/src/PDPVerifier.sol
@@ -290,7 +290,8 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
 
         // Return the at the end to avoid any possible re-entrency issues.
         if (msg.value > sybilFee) {
-            payable(msg.sender).transfer(msg.value - sybilFee);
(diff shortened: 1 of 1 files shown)


aarshkshah1992 (approved):

Stebalien:
> The audit notes that transfers have issues interacting with smart wallet / more sophisticated sending contracts. A large reason for this is that transfer puts a very restricted limit on gas usage in the transfer function. This is a blanket reentrency protection that we don't need. So this PR forwards full gas limit to the transfer to maximize sending contract compatibility.

NOTE: In fevm, we provide 10M gas for transfers so it should be enough in all cases. But we might as well do the right thing.

ZenGround0:
>  In fevm, we provide 10M gas for transfers

Very good to remember

PR #143, opened 2025-04-15
compiler optimizations

ZenGround0:
While we are modifying the mainnet contract in response to issue resolved by #142 we'll also do proper compiler optimizations targeting expensive deploy and cheaper gas running


ZenGround0:
Ok looking closer it does look like this is an optimization of the test code itself which is assuming the mock call can't be hit and so is rewriting the input value from 3 to 5.

ZenGround0:
Ok we're going to try this out

PR #145, opened 2025-04-17
feat: allow provider to set it's beneficiary address in PDP Verifier so all service contracts can get access to it

aarshkshah1992:
Doing this in the Verifier contract allows ALL listeners to access the beneficiary address which they can then use for payments.


ZenGround0 (changes requested):
I don't see this being useful. If the listener wants to specify a separate address from the proofset owner to receive funds that can be done in the extra data.  There's no payments processed in verifier.

PR #150, opened 2025-04-17
Optimize sum-tree operations

Stebalien:
For all I know, `solc` will optimize this for you. But I'm not 100% sure about that.

### DIFF
--- a/src/PDPVerifier.sol
+++ b/src/PDPVerifier.sol
@@ -628,6 +628,8 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
     // Perform sumtree addition
     //
     function sumTreeAdd(uint256 setId, uint256 count, uint256 rootId) internal {
+        mapping(uint256 => uint256) storage set = sumTreeCounts[setId];
+
         uint256 index = rootId;
         uint256 h = heightFromIndex(index);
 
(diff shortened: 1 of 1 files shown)


Stebalien:
This "helps" (~0.3%) with the sum-tree tests, although I'm not sure if we're running the optimizer in CI. It seems to make things worse for the other tests, but I'm not sure if that's noise or if this is somehow increasing the code bloat.

Stebalien:
Honestly, it's likely all noise.

Stebalien:
Ok, the real issue here is that we're using mappings when we should be using dynamic arrays:

- Mappings will require a hash every time we want to look something up.
- Dynamic arrays require a hash to lookup the location of the array, but then we can just index by offset.

We should change `rootLeafCounts`, `sumTreeCOunts` and possibly others to `uint256[]`. TBH, I _think_ we can just set the length of these arrays to "max uint256" if we want them to be sparse, but I'd have to play around with that.

ZenGround0:
Alright this is good information though the timing is unfortunate.  This change would require a state upgrade i.e. a redeploy.  Before jumping into that I'll try to understand the gas costs we're talking about here for the extra keccaks.  If it is a significant fraction i.e. > 20% then we should probably reroll out our contracts.

ZenGround0:
> although I'm not sure if we're running the optimizer in CI

Yes optimizer should be running in CI as of yesterday

Stebalien:
Actually, I think we only need to modify `sumTreeCounts`.

But yeah... I really have no idea what the gas improvements will look like. We'd need to benchmark it on the FVM to get the full picture.

The main reason I'd like to consider this is that hashing currently requires calling out to the FVM so it's not cheap.

Furthermore, it's distributing the `sumTreeCounts` array across the entire contract state which means we're not able to take advantage of data locality.

Stebalien:
> Furthermore, it's distributing the sumTreeCounts array across the entire contract state which means we're not able to take advantage of data locality.

Note: this is very FEVM specific. Basically, reading/writing slots near each other is faster in FEVM than accessing random slots.

ZenGround0:
Ok yeah this definitely seems worth measuring then.  I hadn't been thinking about FEVM coupling when making the mapping choice.

Are you picturing the KAMT data locality savings as dominating this optimization?

Asking because back of the envelope I'm seeing low expected savings just going off of hashing.  fvm charges 33 gas per keccak byte, for uint256 keys mapping hashes 64 bytes (key || slot) so that 2112 gas per hash.  SumTree accesses are log so with a proofset of 1 billion roots ~2^30 we are at ~64k which is around 1 percent of gas costs of the cheapest methods we measure today.

Stebalien:
> Asking because back of the envelope I'm seeing low expected savings just going off of hashing. fvm charges 33 gas per keccak byte, for uint256 keys mapping hashes 64 bytes (key || slot) so that 2112 gas per hash

Unfortunately we also charge 14k gas per syscall (i.e., per hash). We should try to see if we can reduce those costs, but we noticed a significant cost in jumping in and out of the FVM when we first launched it. So it's more like ~480k. But maybe that's still not worth it?

> Are you picturing the KAMT data locality savings as dominating this optimization?

I _think_ it'll let us shave off 4-5 loads when finding a root because our KAMT has a bit-width of 5. Or ~13-15% of the loads. But I'm not sure I'm understanding the tree structure correctly.

Stebalien:
Closing in favor of an issue because my optimizations here are simply not worth it: https://github.com/FilOzone/pdp/issues/153.

ZenGround0:
> So it's more like ~480k. But maybe that's still not worth it?

Got it thanks.  My math above was off in favor of this not mattering -- 60k / 60M is 0.1%.  With 480k we're getting closer to 1 percent but my guess is this is still too small to prioritize.

PR #152, opened 2025-04-17
Use solc 0.8.30

Stebalien:
The MCOPY instruction may provide a significant speedup and I think mainnet should now support everything required by this version.


Stebalien:
Ok, I think we need to update foundry here...

PR #154, opened 2025-04-17
Optimize sumTreeRemove by removing the top check

Stebalien:
If h > top, the index must be passed the end of the array.


Stebalien:
Ah, ok, this is definitely wrong.

Stebalien:
Hm. No, that doesn't matter. The height should be going up by one every time.

Stebalien:
Ooh. The height can increase by _more_ than 1 every time _BECAUSE IT'S NOT A BINARY TREE_ :man_facepalming:.

Stebalien:
Rule: The node's parent is the next node to the right with a greater index.

Stebalien:
This optimization doesn't matter, it's mostly just me trying to remove code we don't need.

PR #156, opened 2025-04-29
init changelog

ZenGround0:


rjan90:
I will see if I can setup some automatic generation of the changelog and release based on a version-string, in line with the other repos we maintain, but we can merge the initial changelog for now.

rjan90 (approved):

PR #160, opened 2025-05-05
Safe migrate function scaffold

### LINKED ISSUE #159: Add versioning
discussed sync w/ @ZenGround0 to add something similar w/  https://github.com/FilOzone/fws-payments/pull/65

ZenGround0:
Closes #159

### DIFF
--- a/src/PDPVerifier.sol
+++ b/src/PDPVerifier.sol
@@ -146,8 +146,7 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
     string public constant VERSION = "1.1.0";
     event ContractUpgraded(string version, address implementation);
 
-    function migrate() public onlyProxy reinitializer(2) {
-        require(msg.sender == address(this), "Only callable by self during upgrade");
+    function migrate() external onlyOwner reinitializer(2) {
         emit ContractUpgraded(VERSION, ERC1967Utils.getImplementation());
     }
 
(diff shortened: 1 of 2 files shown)


src/PDPVerifier.sol
```diff
@@ -143,6 +143,14 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         challengeFinality = _challengeFinality;
     }
 
+    string public constant VERSION = "1.1.0";
+    event ContractUpgraded(string version, address implementation);
+
+    function migrate() public onlyProxy reinitializer(2) {
+        require(msg.sender == address(this), "Only callable by self during upgrade");
```
aarshkshah1992, line 150: Main thing I am still not able to understand is how this works
        require(msg.sender == address(this), "Only callable by self during upgrade");
How can msg.sender  by the same as the proxy addresses if migrate  is essentially called through a delegatecall by the proxy as part of upgradeToAndCall  ?
Shouldn't msg.sender be the EOA that triggered the upgrade ?

src/PDPVerifier.sol
```diff
@@ -143,6 +143,14 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         challengeFinality = _challengeFinality;
     }
 
+    string public constant VERSION = "1.1.0";
+    event ContractUpgraded(string version, address implementation);
+
+    function migrate() public onlyProxy reinitializer(2) {
```
aarshkshah1992, line 149: Why is that `migrate` has an `onlyProxy` modifier but `initialize` does NOT ?

aarshkshah1992 (approved):

PR #161, opened 2025-05-05
Feat/migrate to commpv2

### LINKED ISSUE #158: Use Piece CIDv2
Adin pointed out that PDP isn't using PieceCIDv2, only v1. This is a significant issue because it makes it impossible to fully verify the data from the CID (there's no way to tell how many of the trailing zeros are "padding").

At the moment, we pull the length from the contract, but this defeats many of the benefits of content addressing. The CID should be sufficient by itself.

Luckily, PieceCIDv1 & v2 have identical digests, just different prefixes. So this is mostly an interfaces/display issue. Furthermore, the SP doesn't really care so we can focus on the client interfaces.

ZenGround0:
Closes #158 

Based on #160


jennijuju:
we no longer need backward compatibility if that add complexity

rjan90:
Closing since this has been superseded by: https://github.com/FilOzone/pdp/pull/184

PR #165, opened 2025-05-06
Fallback to unsafe oracle feed when safe call fails

### LINKED ISSUE #164: Cache for oracle failure robustness
@frrist noticed a pdp proving failure on calibration net that has been ongoing for a few days.  The issue is that the Pyth price oracle has not been updated within a day and we require a daily update on pyth oracle from the pdp contract.

To deal with this fallback case we should call `getPriceUnsafe` and emit a warning to the evm logs.  This is in scope for v1.1.0

ZenGround0:
Closes #164 

FYI @frrist

### DIFF
--- a/src/PDPVerifier.sol
+++ b/src/PDPVerifier.sol
@@ -466,7 +466,7 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         }
     }
 
-    function calculateProofFee(uint256 setId, uint256 estimatedGasFee) public view returns (uint256) {
+    function calculateProofFee(uint256 setId, uint256 estimatedGasFee) public returns (uint256) {
         uint256 rawSize = 32 * challengeRange[setId];
         (uint64 filUsdPrice, int32 filUsdPriceExpo) = getFILUSDPrice();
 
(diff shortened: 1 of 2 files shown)


test/PDPVerifier.t.sol
```diff
@@ -1662,6 +1662,38 @@ contract PDPVerifierE2ETest is Test, ProofBuilderHelper {
...
     }
 
+    function createPythUnsafeCallData() internal view returns (bytes memory, PythStructs.Price memory) {
+        bytes memory callData = abi.encodeWithSelector(
+            IPyth.getPriceUnsafe.selector,
+            pdpVerifier.FIL_USD_PRICE_FEED_ID()
+        );
+
+        PythStructs.Price memory price = PythStructs.Price({
+            price: 6,
```
ZenGround0, line 1672: Doing these fixed flag values isn't the best but its consistent with the current state of the world and gets it done.

src/PDPVerifier.sol
```diff
@@ -712,16 +712,24 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
...
-
-        // Return the price and exponent representing USD per FIL
-        return (uint64(priceData.price), priceData.expo);
+        ) returns (PythStructs.Price memory priceData) {
+            require(priceData.price > 0, "failed to validate: price must be greater than 0");
+            // Return the price and exponent representing USD per FIL
+            return (uint64(priceData.price), priceData.expo);
+        }   catch (bytes memory reason) {
+            // Log issue and fallback on latest unsafe price data
+            emit PriceOracleFailure(reason);
```
ZenGround0, line 729: We can listen for this event offchain and in the case of StalePrice automatically refresh the Pyth oracle.  It is a simple api call to get the signed data attesting the price you care about and updating the oracle contract.

aarshkshah1992: @ZenGround0 So that would be a long running service we host ?

aarshkshah1992 (approved):

src/PDPVerifier.sol
ZenGround0: Yes anyone who cares about accurate pdp fee calculations.

PR #169, opened 2025-06-03
Add `pieceCids` to `PiecesAdded` event

juliangruber:
This allows indexers to map roots via their cids, not just their ids. This is required for FilCDN, as root cids will be part of the asset URL.

The migration test is failing. I have no idea why. Could it be because I'm using the `Cids.Cid` type in the event interface?

### DIFF
--- a/src/PDPVerifier.sol
+++ b/src/PDPVerifier.sol
@@ -61,7 +61,7 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
     event DataSetDeleted(uint256 indexed setId, uint256 deletedLeafCount);
     event DataSetEmpty(uint256 indexed setId);
 
-    event PiecesAdded(uint256 indexed setId, uint256[] pieceIds, bytes[] pieceCids);
+    event PiecesAdded(uint256 indexed setId, uint256[] pieceIds, Cids.Cid[] pieceCids);
     event PiecesRemoved(uint256 indexed setId, uint256[] pieceIds);
 
     event ProofFeePaid(uint256 indexed setId, uint256 fee, uint64 price, int32 expo);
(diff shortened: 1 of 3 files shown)


ZenGround0:
Sorry I'm late to this.  I don't fully understand why we need this on PDP Verifier and can't apply this event to the pandora listener contract.  It is definitley possible to do that and it seems like the better place.  Am I mistaken that FILCDN is scoped only to the pandora contract?

ZenGround0 (approved):
@juliangruber this is good to go once you rebase.  AFAIK the only changes you'll need to make there are changing the name `root` to `piece`

ZenGround0:
Oh just saw your comment on the compiler error that's interesting

juliangruber:
Interesting, still the same error

src/PDPVerifier.sol
```diff
@@ -445,12 +445,14 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         require(storageProvider[setId] == msg.sender, "Only the storage provider can add pieces");
         uint256 firstAdded = nextPieceId[setId];
         uint256[] memory pieceIds = new uint256[](pieceData.length);
+        bytes[] memory pieceCidsAdded = new bytes[](pieceData.length);
 
         for (uint256 i = 0; i < nPieces; i++) {
             addOnePiece(setId, i, pieceData[i]);
             pieceIds[i] = firstAdded + i;
+            pieceCidsAdded[i] = pieceCids[setId][firstAdded + i].data;
```
juliangruber, line 453: it is this line that is triggering the compiler error

juliangruber: Commenting out the `optimizer_runs` settings prevents the error

bajtos: How is the `pieceCidsAdded` array different from `pieceData`? Can the event payload use `pieceData` instead?

Issue #173, opened 2025-06-30
Add Root Enumeration Method to PDPVerifier


rvagg:
Currently, PDPVerifier stores all root CIDs for proof sets but provides no efficient way to enumerate them. To get all roots for a proof set, clients must make potentially hundreds of individual RPC calls (`getRootCid()` for each rootId), making it impractical for applications to retrieve complete root lists.

### Proposed Solution

Add a single enumeration method to PDPVerifier that returns all active roots for a proof set, primarily useful via `eth_call`:

```solidity
// Get all active roots for a proof set
function getActiveRoots(uint256 setId) external view returns (
    Cids.Cid[] memory roots,
    uint256[] memory rootIds,
    uint256[] memory rawSizes
);
```

### Implementation Details

The method would:
- Iterate through all rootIds from 0 to `nextRootId[setId] - 1`
- Skip inactive roots (where `rootLeafCounts[setId][rootId] == 0`)
- Return parallel arrays of:
  - `roots`: The CID bytes of each active root
  - `rootIds`: The ID of each active root
  - `rawSizes`: The actual data size in bytes for each root

Depending on how we upgrade to CommPv2, the `rawSize` may be included in the CID bytes so this third return type may not be necessary.

### Example Usage

```js
// Get all roots for a client's proof set
const { roots, rootIds, rawSizes } = await pdpVerifier.getActiveRoots(proofSetId);

// Process results
roots.forEach((root, i) => {
  console.log(`Root ${rootIds[i]}: ${root} (${rawSizes[i]} bytes)`);
});
```

juliangruber:
> Add a single enumeration method to PDPVerifier that returns all active roots for a proof set, primarily useful via eth_call:

There's a risk here: The size of the return value could exceed the maximum transaction size. I suggest to either add something like pagination/offsets, or to continue getting a single one per RPC call. Since these calls can be parallelized heavily, that should still be ok for many cases. It's also what we do [here](https://github.com/checkernetwork/spark-rsr-contract?tab=readme-ov-file#usage), used by https://filecoin-project.github.io/filecoin-storage-providers-market/

rvagg:
```solidity
  // Get active roots for a proof set with pagination
  function getActiveRoots(
      uint256 setId,
      uint256 offset,
      uint256 limit
  ) external view returns (
      Cids.Cid[] memory roots,
      uint256[] memory rootIds,
      uint256[] memory rawSizes,
      uint256 totalActive
  );
```

Where:
[...]

PR #177, opened 2025-07-10
feat: root enumeration with getActiveRoots()

rvagg:
Closes: https://github.com/FilOzone/pdp/issues/173

### DIFF
--- a/src/PDPVerifier.sol
+++ b/src/PDPVerifier.sol
@@ -254,6 +254,69 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         return result;
     }
 
+    /**
+     * @notice Returns active roots (non-zero leaf count) for a proof set with pagination
+     * @param setId The proof set ID
+     * @param offset Starting index for pagination (0-based)
+     * @param limit Maximum number of roots to return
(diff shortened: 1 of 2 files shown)


rvagg:
> I wonder if we can do it avoiding the `totalActive` count

We could, but there'd be trade offs: being able to return a `totalActive` means the caller can decide how to proceed with additional calls. But, we could do a `limit+1` check and return a `hasMore` boolean that serves approximately the same role. This call is intended (by me) to be used by `eth_call`, where gas doesn't matter so much. But if we imagine this to be also useful as a view call then we could be more conscious of gas. We could slash the amount of gas by cutting the number of SLOAD calls in ~half with a single pass and a `hasMore`. We'd then have to wear the cost of a second allocation and copy when we are under `limit` items. There's also the option of storing the total in contract state and keeping that updated as we add/remove roots, but that's more state which is yukky.

Anyway, I've done that so we can see what it looks like. See latest commit 5a78849.

@juliangruber do you have any opinions here? I don't mind just having a `hasMore` boolean. But we could also add a `getActiveRootCount` as a separate function that would give us that info if we needed it.

rvagg:
good points, so I've added `getActiveRootCount`. @Kubuxu would you mind taking another look?

PR #184, opened 2025-08-05
Support PieceCIDv2

Kubuxu:
First commit only right now, supports producing and parsing CIDv2
Removal of rawSize is also going to be there.

Lost quite a bit of time because I updated solidity to 0.8.30 which caused 3rd party failures.

### DIFF
--- a/src/Cids.sol
+++ b/src/Cids.sol
@@ -25,13 +25,14 @@ library Cids {
         return bytes32(dataSlice);
     }
 
+    // Returns the height of the tree from the CID.
     function heightFromCid(Cid memory cid) internal pure returns (uint8) {
         require(cid.data.length >= 33, "Cid data is too short");
         return uint8(cid.data[cid.data.length - 32 - 1]);
     }
 
 
-    // Checks that cid matches CommPv2.
+    // Checks that CID is CommPv2 and decomposes it into its components.
     // See: https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0069.md
     function validateCommPv2(Cid memory cid)
         internal
@@ -58,6 +59,7 @@ library Cids {
         return (padding, height, offset);
     }
 
+    // isPaddingExcessive checks if the padding size exceeds the size of the tree
     function isPaddingExcessive(uint256 padding, uint8 height) internal pure returns (bool) {
         return (128*padding)/127 >= 1<<(height+5);
     }
@@ -81,8 +83,6 @@ library Cids {
         // which then after subtracting the padding leaves is the number of leaves that contain data
         return (1 << uint256(height)) - paddingLeafs;
     }
-
-
 
     // Creates a CommPv2 CID from a raw size and hash digest according to FRC-0069.
     // The CID uses the Raw codec and fr32-sha2-256-trunc254-padded-binary-tree multihash.


Kubuxu:
@ZenGround0 this is ready for initial review. I will add more tests, but I have managed to get every existing test to work.

src/Cids.sol
```diff
@@ -1,8 +1,16 @@
-// SPDX-License-Identifier: Apache-2.0 OR MIT
+// SPDX-License-Identifier: UNLICENSED
```
rvagg, line 1: BRO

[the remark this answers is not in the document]
Kubuxu: No, that was in your Cid file, I just haven't noticed it (I did `git checkout $yourbranch -- src/Cids.sol`

src/Cids.sol
```diff
@@ -1,8 +1,16 @@
...
 pragma solidity ^0.8.20;
 
+import {BitOps} from "./BitOps.sol";
+import "@openzeppelin/contracts/utils/Strings.sol";
+
 library Cids {
-    // TODO PERF: https://github.com/FILCAT/pdp/issues/16#issuecomment-2329836995
+    uint256 public constant COMMP_LEAF_SIZE = 32;
+    //  0x01    0x55                0x9120                              0x21
```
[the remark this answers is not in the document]
Kubuxu: yeah, I started adding it the prefix but then removed, going to remove it from the comment

src/Cids.sol
```diff
@@ -11,22 +19,122 @@ library Cids {
...
+        (padding, offset) = _readUvarint(cid.data, offset);
+
+        height = uint8(cid.data[offset]);
+        offset++;
+
+        return (padding, height, offset);
+    }
+
+    function isPaddingExcessive(uint256 padding, uint8 height) internal pure returns (bool) {
+        return (128*padding)/127 >= 1<<(height+5);
```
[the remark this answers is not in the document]
Kubuxu: This is a check to catch the case where there are zero data leaves.

src/Cids.sol
```diff
@@ -11,22 +19,122 @@ library Cids {
...
+        uint256 offset = 4;
+        uint256 mhLength;
+        (mhLength, offset) = _readUvarint(cid.data, offset);
+        require(mhLength >= 34, "CommPv2 multihash length must be at least 34");
+        if (mhLength + offset != cid.data.length) {
+            revert("CommPv2 multihash length does not match data length");
         }
+        (padding, offset) = _readUvarint(cid.data, offset);
+
+        height = uint8(cid.data[offset]);
```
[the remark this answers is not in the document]
Kubuxu: I don't think it is required per se, and I don't think the spec disallows over-padded CommPs (cases where data could be covered by smaller tree with less padding).
These might not be canonical, and CommPv2 producers should avoid creating these cases, but they are not invalid.
Yes, it leads to the same piece of data having multiple different CID references, but as long as we recognise the minimal one as canonical and recommend everyone use it, it should be fine.
What do you think?

src/Cids.sol
Kubuxu: accidental left-over from me checking out this file from another branch, going to fix

src/Cids.sol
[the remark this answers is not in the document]
rvagg: docs are good for this kind of thing, remember that this is all basically esoteric knowledge

src/Cids.sol
```diff
@@ -11,22 +19,133 @@ library Cids {
...
+        // so we need to expand it by factor 128/127
+        // then we divide by 32 with a floor to get the number of leaves that are fully padding
+        uint256 paddingLeafs = (128*padding)/127 >> 5;
+        // 1<<height is the number of leaves in the tree
+        // wihch then after subtracting the padding leaves is the number of leaves that contain data
+        return (1 << uint256(height)) - paddingLeafs;
+    }
+
+
+
```
rvagg, line 85:
```suggestion

```

src/Cids.sol
```diff
@@ -11,22 +19,133 @@ library Cids {
...
+    }
+
+    // leafCount returns the number of 32b leaves that contain any amount of data
+    function leafCount(uint256 padding, uint8 height) internal pure returns (uint256) {
+        // the padding itself is # of bytes before Fr32 expansion
+        // so we need to expand it by factor 128/127
+        // then we divide by 32 with a floor to get the number of leaves that are fully padding
+        uint256 paddingLeafs = (128*padding)/127 >> 5;
+        // 1<<height is the number of leaves in the tree
+        // wihch then after subtracting the padding leaves is the number of leaves that contain data
```
rvagg, line 80:
```suggestion
        // which then after subtracting the padding leaves is the number of leaves that contain data
```

test/Cids.t.sol
```diff
@@ -43,4 +43,96 @@ contract CidsTest is Test {
...
+        // Write all values
+        for (uint256 i = 0; i < values.length; i++) {
+            offset = Cids._writeUvarint(buffer, offset, values[i]);
+        }
+
+        // Read all values and verify
+        uint256 currentOffset = 0;
+        for (uint256 i = 0; i < values.length; i++) {
+            (uint256 readValue, uint256 newOffset) = Cids._readUvarint(buffer, currentOffset);
+            assertEq(readValue, values[i], "Uvarint round trip failed");
```
ZenGround0, line 85: Maybe writing everything to one buffer and then reading it out again is a common pattern but I don't think I've seen it before.  It is a nice way to exercise the offset return as well.

test/Cids.t.sol
```diff
@@ -43,4 +43,96 @@ contract CidsTest is Test {
...
+        Cids._readUvarint(incompleteUvarint2, 0);
+
+        // Test reading with an offset
+        bytes memory bufferWithOffset = hex"00010203040506078001"; // Value 128 (8001) at offset 8
+        (uint256 readValue, uint256 newOffset) = Cids._readUvarint(bufferWithOffset, 8);
+        assertEq(readValue, 128, "Read uvarint with offset failed");
+        assertEq(newOffset, 10, "Offset after reading with offset incorrect");
+    }
+
+    function testValidateCommPv2Golden() public pure {
```
ZenGround0, line 108: Naming means this is golden file style but to be clear we are not meaning to be checking in an actual golden file right?

test/Cids.t.sol
```diff
@@ -43,4 +43,96 @@ contract CidsTest is Test {
...
+            currentOffset = newOffset;
+        }
+    }
+
+    /// forge-config: default.allow_internal_expect_revert = true
+    function testReadUvarintEdgeCases() public {
+        // Test reading an incomplete uvarint that should revert
+        bytes memory incompleteUvarint = hex"80"; // A single byte indicating more to come, but nothing follows
+        vm.expectRevert(); // Expect any revert, specifically index out of bounds
+        Cids._readUvarint(incompleteUvarint, 0);
```
ZenGround0, line 95: This function exits after this call.  Expect revert doesn't catch the error and keep going, it catches it stops the revert and then returns.  So the code below is not run.

test/Cids.t.sol
```diff
@@ -43,4 +43,96 @@ contract CidsTest is Test {
...
+
+        bytes memory incompleteUvarint2 = hex"ff01"; // MSB set on last byte.
```
ZenGround0, line 97: This is the valid encoding of 255. So this will fail once you resolve the issue above.  You probably meant `ff81`?

rjan90 (approved):

PR #186, opened 2025-08-21
fix(IPDPProvingSchedule): Mark View Methods

wjmelements:
Reviewer @zenground0
I want FilecoinWarmStorageServiceStateView to implement IPDPProvingSchedule.
Most of these methods are actually view methods right now, mostly because these fields are in storage rather than being constants:
* `getMaxProvingPeriod`
* `challengeWindow`

I would prefer to change FilecoinWarmStorageService to make these constants, but it looks like they are intended to be configurable. Therefore these methods should be `view` instead of `pure`.

#### Changes
* Fix interface to actual view/pure in current FilecoinWarmStorageService

### DIFF
--- a/src/IPDPProvingSchedule.sol
+++ b/src/IPDPProvingSchedule.sol
@@ -6,15 +6,15 @@ pragma solidity ^0.8.20;
 interface IPDPProvingSchedule {
     /// @notice Returns the number of epochs allowed before challenges must be resampled
     /// @return Maximum proving period in epochs
-    function getMaxProvingPeriod() external pure returns (uint64);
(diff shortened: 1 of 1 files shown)


wjmelements:
> > I would prefer to change FilecoinWarmStorageService to make these constants.
> > I will also create a PR for that, though it is a more complex change.
> 
> It is not hard to make them constants but It is hard for me to determine what they should be.

It's also apparent that they're supposed to be configurable, so these methods should be `view` rather than `pure`. Going to suspend work on the `pure` constant approach for now, favoring this PR.

PR #191, opened 2025-08-27
Pyth: accept old prices without emit log

wjmelements:
Reviewer @zenground0

Curio is expecting `calculateProofFee` to be `view`.
What is preventing that is that we are logging that the timestamp is out of date before just using the data anyway.
That isn't a great way to handle the situation.
The prior behavior is equivalent to `getPriceUnsafe` but with an extra staticcall in the outdated case.

If we want to notice that the price is outdated we can monitor the oracle directly; this is no more difficult than watching for `PriceOracleFailure` logs.
On-chain logs shouldn't be used like `logger` logs.
A better way to think of them is like a SQL database row.

It's also probably not a great experience for the price to change and then your transaction reverts, but that's a separate issue.
#### Changes
* use `getPriceUnsafe`, remove log
* make `calculateProofFee` a `view` method
* fix tests

### DIFF
--- a/src/PDPVerifier.sol
+++ b/src/PDPVerifier.sol
@@ -577,7 +577,7 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         }
     }
 
-    function calculateProofFee(uint256 setId, uint256 estimatedGasFee) public returns (uint256) {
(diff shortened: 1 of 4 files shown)


ZenGround0 (approved):

ZenGround0:
Good change. Should have just done it this way in the first place.

PR #195, opened 2025-09-15
feat: add automated draft release on version change

rjan90:
Add automated release workflow for version changes:
- Detects version changes and creates draft releases with auto-generated changelog
- Automatically publishes draft releases when PRs are merged to main

### Workflow:
- PR with version change → Creates draft release with changelog
- Review/edit draft release as needed
- Merge to main → Automatically publishes the release

### DIFF
--- a/.github/workflows/version-release.yml
+++ b/.github/workflows/version-release.yml
@@ -1,0 +1,275 @@
+name: Auto Release on Version Change
+
+# Automatically creates and publishes releases when the VERSION constant in PDPVerifier.sol changes
+# This workflow:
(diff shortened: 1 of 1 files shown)


rvagg:
I wonder if https://github.com/googleapis/release-please might just be a better option here. I'm going with that in filecoin-pin at the moment and we'll be switching synapse-sdk to it and it's fairly flexible, not tied to npm/node stuff. Maybe we can extend it with some hooks to do some updates for us - like update the version string in the contract for us based on conventional commits guessing. Worth looking at at least?

PR #196, opened 2025-09-16
feat: start data set IDs at 1

rvagg:
Avoid problems with default value matching. Matches how we deal with rail IDs and service provider IDs.

Closes: https://github.com/FilOzone/filecoin-services/issues/216

This should be a non-breaking change to a deployed contract, upgrading would essentially be a noop for any contract that has data sets already (also for any that don't because this is set in initialize()). So it's only relevant for new deployments.


ZenGround0 (approved):
Thanks, should have done this in the first place

PR #200, opened 2025-09-18
fix: default to dry-run mode with upgrade-script

rjan90:
Default to dry-run mode with `upgrade-contract-calibnet.sh` script

### DIFF
--- a/tools/upgrade-contract-calibnet.sh
+++ b/tools/upgrade-contract-calibnet.sh
@@ -5,6 +5,15 @@
 # and to a valid RPC_URL for the devnet.
 # Assumption: forge, cast, jq are in the PATH
 #
+# Set DRY_RUN=false to actually deploy and broadcast transactions (default is dry-run for safety)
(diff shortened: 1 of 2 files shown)


rvagg (approved):
nice, while you're in here would you mind upgrading the chain ID based on the model in the other scripts, something like:

```sh
if [ -z "$CHAIN_ID" ]; then
  CHAIN_ID=$(cast chain-id --rpc-url "$RPC_URL")
  if [ -z "$CHAIN_ID" ]; then
    echo "Error: Failed to detect chain ID from RPC"
    exit 1
  fi
fi
```

then use that rather than a magic number, it makes it more portable too

rjan90:
Refactored in: https://github.com/FilOzone/pdp/pull/200/commits/2150b92274601c4d512845d0efed68e2b868a1f2
- Made it network agnostic, and figure out the chain detection via `cast chain-id --rpc-url`
- Renamed from `upgrade-contract-calibnet.sh` to `upgrade-contract.sh` to reflect this.

PR #201, opened 2025-09-18
feat!: Merge createDataset and addPieces for streamlined dataset creation

Chaitu-Tatipamula:
## Description

This PR implements the feature described in [#171 at filecoin-services](https://github.com/FilOzone/filecoin-services/issues/171) to merge `createDataset` and `addPieces` operations for a streamlined user experience.

### PDP Verifier Updates
- **Enhanced `addPieces` function**: Now accepts `(datasetID, listenerAddress, pieces, extraData)` parameters
- **New dataset creation**: Use `datasetID = NEW_DATA_SET_SENTINEL` (type(uint256).max) with a valid `listenerAddress` to create a new dataset and add pieces in a single transaction
- **Existing dataset updates**: Use `listenerAddress = 0` with an existing `datasetID` to add pieces to existing datasets
- **Removed `createDataset` method**: No longer needed as functionality is merged into `addPieces`

### Test Improvements
- Added comprehensive test coverage for the combined create-and-add flow
- Fixed listener initialization issues across test contracts
- Added `TestConstants.sol` for shared constants like `NEW_DATA_SET_SENTINEL`
- Updated all test contracts to use the new streamlined approach

### Benefits
- ✅ **Single transaction**: Users can create datasets and add pieces in one operation
- ✅ **Reduced wait times**: No need to wait for dataset creation confirmation before adding pieces
- ✅ **Simplified UX**: One client signing operation instead of two
- ✅ **Backward compatibility**: Existing functionality for adding to existing datasets preserved

## Testing

- All existing tests pass
- New tests cover the combined create-and-add flow
- Listener initialization issues resolved
- Event emission verified for both dataset creation and piece addition

## Related Issue

Closes  [#171 at filecoin-services](https://github.com/FilOzone/filecoin-services/issues/171)

PR #219, opened 2025-10-08
feat: restore createDataSet

rvagg:
Removed in https://github.com/FilOzone/pdp/pull/201 as per original request. This puts that function back in just to give us the option to create empty data sets if needed, and gives us a smoother transition path for our dependencies (mainly Curio).

---

I'm just putting this option on the table, it doesn't have to be committed to, but it will make life a bit easier.

Ultimately the goal is to not need or use this, but it's not terrible having the option to create empty data sets, it solves the desire to create empty "buckets" S3 style, as mentioned in the original issue https://github.com/FilOzone/filecoin-services/issues/171 (and also expressed separately by Hugo).

See:
* https://github.com/filecoin-project/curio/issues/678
* https://github.com/filecoin-project/curio/pull/687
* https://github.com/filecoin-project/curio/pull/685 -- this is probably the closest to getting it working
* https://github.com/FilOzone/synapse-sdk/pull/287

### DIFF
--- a/src/PDPVerifier.sol
+++ b/src/PDPVerifier.sol
@@ -177,6 +177,23 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         require(success, "Burn failed");
     }
 
+    // Validates msg.value meets sybil fee requirement and burns the fee.
+    // Returns the sybil fee amount for later refund calculation.
+    function _validateAndBurnSybilFee() internal returns (uint256 sybilFee) {
+        sybilFee = PDPFees.sybilFee();
+        require(msg.value >= sybilFee, "sybil fee not met");
+        burnFee(sybilFee);
+    }
+
+    // Refunds any amount sent over the sybil fee back to msg.sender.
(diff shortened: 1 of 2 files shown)


src/PDPVerifier.sol
```diff
@@ -408,9 +464,28 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         emit DataSetDeleted(setId, deletedLeafCount);
     }
 
-    // Create Dataset and Add Pieces, When setId == NEW_DATA_SET_SENTINEL, this will create a new dataset with piece data provided
-    // with the provided listenerAddr and expect extraData to be abi.encode(bytes createPayload, bytes addPayload).
-    // When adding to an existing set, pass listenerAddr == address(0) and setId to the live dataset.
+    // Appends pieces to a data set. Optionally creates a new data set if setId == 0.
+    // These pieces won't be challenged until the next proving period is started by calling nextProvingPeriod.
```
ZenGround0, line 468: 🙏

ZenGround0 (approved):
Love it

PR #228, opened 2025-10-23
fix: prevent duplicate piece IDs in schedulePieceDeletions

### LINKED ISSUE #227: `schedulePieceDeletions` doesn't check for duplicate piece IDs
## Description
The schedulePieceDeletions function in PDPVerifier.sol does not check for duplicate piece IDs, allowing the same piece to be scheduled for removal multiple times. This can happen in two ways:
1. Calling schedulePieceDeletions multiple times with the same piece ID
2. Passing duplicate piece IDs within a single schedulePieceDeletions call

https://github.com/FilOzone/pdp/blob/937b93ac8966b32b60215c51be38f73c2f498a80/src/PDPVerifier.sol#L568-L572

## Steps to Reproduce

1. Call schedulePieceDeletions with piece ID 0 on dataset 9
2. Call schedulePieceDeletions again with the same piece ID 0 on dataset 9
3. Both transactions succeed, and piece ID 0 is now scheduled for removal twice

Calibnet testnet transaction showing successful duplicate deletion:
https://filecoin-testnet.blockscout.com/tx/0x9e3311da7f442b039bbc05bbfeac72a18187702d49d745457f84e5885b23a80a

Chaitu-Tatipamula:
### Problem
The `schedulePieceDeletions` function didn't check for duplicate piece IDs, allowing the same piece to be scheduled for removal multiple times. This could happen within a single call or across multiple calls.

### Changes
- Added `isScheduledForRemoval` mapping for O(1) lookups for duplicate checking
- Updated `schedulePieceDeletions` to validate against existing scheduled removals
- Added mapping cleanup in `nextProvingPeriod` to prevent stale state
- Added comprehensive test coverage for duplicate prevention

### Tests
- `testSchedulePieceDeletionsDuplicatePrevention`: Verifies duplicate detection works for both within-call and cross-call scenarios
- `testMappingClearedAfterRemoval`: Ensures mapping is properly cleared after removals are processed
- All existing tests continue to pass

Fixes #227

### DIFF
--- a/src/PDPVerifier.sol
+++ b/src/PDPVerifier.sol
@@ -549,13 +549,8 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         }
 
         uint256 leafCount = Cids.leafCount(padding, height);
-        if (leafCount == 0) {
-            // This is a redudant check as Cids.isPaddingExcessive already checks for pieces which would result in leafCount == 0
-            // but we keep it here for clarity
-            revert IndexedError(callIdx, "Leaf count of the pieece must be greater than 0");
-        }
-
         uint256 pieceId = nextPieceId[setId]++;
+
         sumTreeAdd(setId, leafCount, pieceId);
         pieceCids[setId][pieceId] = piece;
         pieceLeafCounts[setId][pieceId] = leafCount;
--- a/src/Cids.sol
+++ b/src/Cids.sol
@@ -55,6 +55,7 @@ library Cids {
     }
 
     // isPaddingExcessive checks if the padding size exceeds the size of the tree
+    // When this is false, leafCount will never return zero
     function isPaddingExcessive(uint256 padding, uint8 height) internal pure returns (bool) {
         return (128 * padding) / 127 >= 1 << (height + 5);
     }
@@ -69,6 +70,8 @@ library Cids {
     }
 
     // leafCount returns the number of 32b leaves that contain any amount of data
+    // Utilize isPaddingExcessive to check if the padding size exceeds the size of the tree
+    // If isPaddingExcessive is false, leafCount will never return a zero.
     function leafCount(uint256 padding, uint8 height) internal pure returns (uint256) {
         // the padding itself is # of bytes before Fr32 expansion
         // so we need to expand it by factor 128/127


src/PDPVerifier.sol
```diff
@@ -130,6 +130,8 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
     mapping(uint256 => uint256) challengeRange;
     // Enqueued piece ids for removal when starting the next proving period
     mapping(uint256 => uint256[]) scheduledRemovals;
+    // Track which pieces are scheduled for removal
+    mapping(uint256 dataSetId => mapping(uint256 pieceId => bool isScheduledForRemoval)) isScheduledForRemoval;
```
wjmelements: Because pieceId is contiguous, use a uint256 bitmap, like https://github.com/FilOzone/filecoin-services/pull/258

This will reduce the storage access in most cases when there are multiple pieces.

Chaitu-Tatipamula: i was thinking what if there are more than 256 pieces or is it unlikely to have that much? or maybe user can split into multiple datasets?

src/PDPVerifier.sol
```diff
@@ -130,6 +130,8 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
     mapping(uint256 => uint256) challengeRange;
     // Enqueued piece ids for removal when starting the next proving period
     mapping(uint256 => uint256[]) scheduledRemovals;
+    // Track which pieces are scheduled for removal with a bitmap
+    mapping(uint256 dataSetId => uint256 scheduledRemovalsBitmap) scheduledRemovalsBitmap;
```
wjmelements: it should still be a two level map because there can be more than 256 pieces. Shift the pieceId down 8 bits for the key and use the bottom 8 bits to index the uint256 bitmap

Kubuxu (approved):
SGWM, but would like a second set of eyes from @wjmelements given that I pushed some code into this branch

wjmelements (approved):

src/PDPVerifier.sol
```diff
@@ -547,6 +549,12 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         }
 
         uint256 leafCount = Cids.leafCount(padding, height);
+        if (leafCount == 0) {
+            // This is a redudant check as Cids.isPaddingExcessive already checks for pieces which would result in leafCount == 0
+            // but we keep it here for clarity
+            revert IndexedError(callIdx, "Leaf count of the pieece must be greater than 0");
+        }
```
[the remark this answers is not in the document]
Kubuxu: I've added it because I spent a good 15 minutes verifying that it is actually true.

PR #241, opened 2026-01-08
docs(pdp): add per-piece security guarantees section

rvagg:
Also tagging @lucaniz for review here; it's based off https://www.notion.so/filecoindev/PDP-security-and-ProofSet-size-28edc41950c180d088dee430e1249a2a but is user-focused: _"My specific piece wasn't challenged in the last X days—how do I know it's still safe?"_

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -199,6 +199,16 @@ As shown in the table above, detection confidence depends on the fraction of dat
 
 This is fundamentally different from per-piece proving (where each piece would need individual challenges) and is more efficient while providing strong security guarantees for detecting any meaningful data loss.
 
+**Using detection history for trust decisions:**
+
+PDP provides detection confidence, not failure prevention. It answers "if data is lost, how likely are we to catch it?" rather than "will data be lost?" However, a provider's historical proof record serves as a practical indicator of operational reliability. A provider that has successfully proven a data set for 30+ days demonstrates:
+
+1. Functional storage infrastructure
+2. Operational consistency
+3. No detected data loss during that period
+
+A clean proof record is strong evidence of operational reliability, though not a guarantee of future performance.
+
 ### Completeness
 - Proving always works if providing Merkle proofs to the randomly sampled leaves
 


docs/design.md
```diff
@@ -153,6 +153,52 @@ Detailed description of key workflows.
 - Merkle proofs must be sound
 - Randomness cannot be biased through grinding or chain forking
 
+### Per-Piece Security Guarantees
+
+A common concern is: "My specific piece wasn't challenged in the last X days—how do I know it's still safe?"
+
+The key insight is that **successful data set proofs provide strong statistical guarantees for all pieces in the data set**, regardless of which specific pieces were challenged. Random challenge selection means any missing data will be detected with high probability over time.
```
lucaniz: probabilistic

docs/design.md
```diff
@@ -153,6 +153,52 @@ Detailed description of key workflows.
...
+
+**Detection probability over time:**
+
+With one proof per day containing K challenges, the evasion probability after T days is:
+
+```
+p_T = (1-α)^(K×T)
+```
+
+**Example detection rates (K=5 challenges per day):**
```
timfong888: For me, let's say we want a commitment that AFR-1 = 99%.  How do we communicate this?

A customer has one data set and that's where they continue to feed all of their data (say files), and get to 1000 files.
Withint a year, we want to say they will only lose 1 file.  What is the DealBot threshold to know that SLA can be reached?

docs/design.md
```diff
@@ -153,6 +153,52 @@ Detailed description of key workflows.
...
+
+| Data Lost (α) | Daily Detection | 30-Day Detection |
+|---------------|-----------------|------------------|
+| 1%            | 4.9%            | 77.9%            |
+| 5%            | 22.6%           | 99.95%           |
+| 20%           | 67.2%           | ~100%            |
+
+**What this means for individual pieces:**
+
+If a storage provider has lost any significant fraction of a data set, they will be caught with high probability regardless of which specific pieces are missing. The random challenge selection ensures that:
```
timfong888: Here's where I could use more clarity using the use case.  "significant fraction" will be caught.  I am using the 1% because that is the AFR we are using with as a base case.  meaning the Annual Failure Rate is 1%.  We want to approve an SP based on a period of time/number of successful proofs.  How do we reason this on the customer's behalf?

docs/design.md
rvagg: According to the numbers, apparently for 1 file lost out of 1000 (α = 0.1%):
* Daily detection:  1 - (0.999)^5   ≈ 0.5%
* 30-day detection: 1 - (0.999)^150 ≈ 14%
* Annual detection: 1 - (0.999)^1825 ≈ 84%

PR #242, opened 2026-01-11
refactor: follow CEI pattern in `provePossession`

### LINKED ISSUE #238: Move dataSetLastProvenEpoch mutation above external call in provePossession
Reported by @relotnek. As far as we both can see this is not critical, particularly with the gated listener deployment set up that we use exclusively with PDPVerifier today, but it would probably be good to tidy this up by the time we want to do a PDPVerifier v2 deployment.

https://github.com/FilOzone/pdp/blob/b138bd534f760d7add1ae284f4ae47142f489138/src/PDPVerifier.sol#L677-L681

`dataSetLastProvenEpoch` mutation happens after the call out to the listener, violating the checks-effects-interactions pattern, potentially allowing weird reentrancy problems. It probably should be done before that call. Most interesting calls are guarded with `storageProvider[setId] == msg.sender` so it seems hard to imagine a set up where a malicious listener could get involved here, particularly since the SP is in charge of gating who the listener is.

Recording this so it doesn't get lost, nice tidy-up at some point later, I'm not suggesting we do this now because current code represents v1 and there's currently no pressure to get a new one out.

akronim26:
Fixes #238

### DIFF
--- a/src/PDPVerifier.sol
+++ b/src/PDPVerifier.sol
@@ -670,6 +670,7 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         //
         // (add 32 bytes to the `callDataSize` to also account for the `setId` calldata param)
         uint256 refund = calculateAndBurnProofFee(setId);
+        dataSetLastProvenEpoch[setId] = block.number;
 
         {
             address listenerAddr = dataSetListener[setId];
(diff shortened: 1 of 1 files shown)


rvagg:
this is fine, but the current head commit in the main branch represents the v1 deployment so we might just put this on hold for now until we're a bit more clear of our GA release

wjmelements (approved):

PR #245, opened 2026-01-29
feat: add tool to get initialization counter and update that in contr…

### LINKED ISSUE #244: `immutable` reinitializer parameter
FS issue https://github.com/FilOzone/filecoin-services/issues/401
FS PR https://github.com/FilOzone/filecoin-services/pull/402

> To avoid future mistakes where we forget to update the initializer count, we can generate it at deployment time after fetching the previous value with eth_getStorageAt.

DarkLord017:
Fixes #244 

introduces a new upgrade pattern for the `PDPVerifier` and `SimplePDPService` contracts by adding a reinitializer version to be set via the constructor. 

All deployment scripts and tests have been updated to pass the required constructor argument.

### DIFF
--- a/test/PDPVerifier.t.sol
+++ b/test/PDPVerifier.t.sol
@@ -1182,8 +1182,7 @@ contract TestingRecordKeeperService is PDPListener, PDPRecordKeeper {
 }
 
 contract SumTreeInternalTestPDPVerifier is PDPVerifier {
-    constructor() PDPVerifier(1) {
-    }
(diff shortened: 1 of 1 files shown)


DarkLord017:
> Will need a follow-on PR to filecoin-service to pull this submodule and update the scripts there that deploy the PDPVerifier implementation (such as `tools/deploy-all-warm-storage.sh`).

Okay will do this

PR #262, opened 2026-03-24
feat: add scripts to gen and verify storage layout

### LINKED ISSUE #258: Setup `*ServiceLayout.sol` and enforce that we only make additions
## Done Criteria
This upgradable contract has:
1. A `*ServiceLayout.sol` file that is generated from the contract source code as part of `mage gen`
2. A CI job that makes sure that the checked-in `*ServiceLayout.sol` file matches what is freshly generated.
3. A CI check that ensures `*ServiceLayout.sol` only has storage slot additions (i.e., increasing numbers, not overriding or changing existing numbers).
4. Release process with an explicit step for human checking that `*ServiceLayout.sol` has no destructive changes between the current contract release and the to-be-released contract.

## Why Important
Same as https://github.com/FilOzone/filecoin-services/issues/446
 
## User/Customer
Maintainers

## Notes
1. This issue is effectively about getting PdpVerifier to have the same storage slot operational safety that `filecoin-services` has after https://github.com/FilOzone/filecoin-services/issues/446 is completed. We should be able to draw a lot from that repo.

akronim26:
Fixes #258


rjan90 (changes requested):
There is quite a bit of unnecessary changes that are formatting-only edits in. Please drop the changes in src/PDPVerifier.sol, src/SimplePDPService.sol, test/PDPVerifier.t.sol and test/SimplePDPService.t.sol.

Additionally I think you can remove the unrelated lib/pyth-sdk-solidity submodule bump.

PR #263, opened 2026-03-24
  feat: enforce additive-only storage layout changes for PDPVerifier

### LINKED ISSUE #258: Setup `*ServiceLayout.sol` and enforce that we only make additions

Chaitu-Tatipamula:
## Summary
  Implements storage slot safety for the PDPVerifier upgradable contract
  by enforcing that storage layout changes are strictly additive,
  preventing storage collision bugs during contract upgrades.

  Resolves #258

## What's Checked                                                       
  | Scenario | Detected |
  |----------|----------|                                                 
  | ❌ Removing existing storage slots | Yes |                            
  | ❌ Changing the slot number of an existing variable | Yes |           
  | ❌ Inserting new slots in the middle (shifting existing slots) | Yes |
  | ✅ Appending new slots at the end | Allowed |          
  ## Changes
  - **tools/generate_storage_layout.sh**: Script to generate
  `PDPVerifierLayout.sol` from contract source using `forge inspect`
  - **tools/check_storage_layout.sh**: Validates that storage layout
  changes only add new slots at the end; flags destructive changes like
  removed slots, moved slots, or inserted slots
  - **src/PDPVerifierLayout.sol**: Generated file with 17 storage slots
  (must be committed and kept in sync)
  - **Makefile**: Added `gen`, `check-layout`, `clean-gen` targets for
  managing storage layout
  - **.github/workflows/check-storage-layout.yml**: New CI workflow to
  verify storage layout on all pushes/PRs
  - **.github/workflows/makefile.yml**: Added storage layout check to
  existing CI

  ## Usage
  ```bash
  make gen              # Generate/rebuild storage layout
  make check-layout     # Check for destructive changes (runs in CI)
  make clean-gen        # Clean generated files       
  ```

### DIFF
--- a/.github/workflows/makefile.yml
+++ b/.github/workflows/makefile.yml
@@ -34,3 +34,8 @@ jobs:
         run: |
           export PATH="/home/runner/.config/.foundry/bin:$PATH";
           make test;
+
+      - name: Check storage layout
+        run: |
+          export PATH="/home/runner/.config/.foundry/bin:$PATH";
+          make check-layout;
--- a/.github/workflows/makefile.yml
+++ b/.github/workflows/makefile.yml
@@ -12,6 +12,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
 
       - name: configure
         run: ./configure
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ $(LAYOUT): tools/generate_storage_layout.sh src/PDPVerifier.sol
 
 # Storage layout JSON (full metadata for upgrade safety checks)
 $(LAYOUT_JSON): src/PDPVerifier.sol
-	forge inspect --json src/PDPVerifier.sol:PDPVerifier storageLayout | jq '[.storage[] | {label, slot, offset, type}]' > $@
+	forge inspect --json src/PDPVerifier.sol:PDPVerifier storageLayout | jq '[.types as $$types | .storage[] | {label, slot, offset, type: ($$types[.type].label // .type)}]' > $@
 
 # Main code generation target
 .PHONY: gen
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ PASSWORD ?=
 
 # Generated files
 LAYOUT=src/PDPVerifierLayout.sol
+LAYOUT_JSON=src/PDPVerifierLayout.json
 
 # Default target
 .PHONY: default
@@ -63,9 +64,13 @@ contract-size-check:
 $(LAYOUT): tools/generate_storage_layout.sh src/PDPVerifier.sol
 	bash tools/generate_storage_layout.sh src/PDPVerifier.sol:PDPVerifier | forge fmt -r - > $@
 
+# Storage layout JSON (full metadata for upgrade safety checks)
+$(LAYOUT_JSON): src/PDPVerifier.sol
+	forge inspect --json src/PDPVerifier.sol:PDPVerifier storageLayout | jq '[.storage[] | {label, slot, offset, type}]' > $@
+
 # Main code generation target
 .PHONY: gen
-gen: check-tools $(LAYOUT)
+gen: check-tools $(LAYOUT) $(LAYOUT_JSON)
 	@echo "Code generation complete"
 
 # Force regeneration - useful when things are broken
@@ -77,7 +82,7 @@ force-gen: clean-gen gen
 .PHONY: clean-gen
 clean-gen:
 	@echo "Removing generated files..."
-	@rm -f $(LAYOUT)
+	@rm -f $(LAYOUT) $(LAYOUT_JSON)
 	@echo "Generated files removed"
 
 # Check required tools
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -54,4 +57,37 @@ extract-abis:
...
+.PHONY: check-tools
+check-tools:
+	@which jq >/dev/null 2>&1 || (echo "Error: jq is required but not installed" && exit 1)
+	@which forge >/dev/null 2>&1 || (echo "Error: forge is required but not installed" && exit 1)
+
+# Storage layout validation
+.PHONY: check-layout
+check-layout:
+	@echo "Checking storage layout for destructive changes..."
+	@bash tools/check_storage_layout.sh
```
rjan90: `check-layout` only validates the checked-in `PDPVerifierLayout.sol`; it never regenerates it from `src/PDPVerifier.sol`. That means CI can pass with a stale layout file. Please make this check rebuild the layout first and fail if the regenerated output differs from the committed file, then run the additive-only validation.

tools/check_storage_layout.sh
```diff
@@ -0,0 +1,209 @@
...
+        TEMP_BASE_LAYOUT=$(mktemp)
+        TEMP_FILES+=("$TEMP_BASE_LAYOUT")
+
+        if ! git show "$BASE_REF:$FULL_LAYOUT_FILE" > "$TEMP_BASE_LAYOUT" 2>/dev/null; then
+            echo "Warning: Could not retrieve base layout, assuming new file"
+            if validate_layout_format "$LAYOUT_FILE"; then
+                echo "Storage layout format validated"
+                exit 0
+            else
+                exit 1
```
rjan90: This fallback is too permissive for CI: when the base ref or base layout cannot be loaded, the script warns and exits successfully after format-only validation. That downgrades the safety check to a no-op. Please make missing base refs a hard failure in CI, and only skip comparison when the layout file truly does not exist yet on the base branch.

.github/workflows/makefile.yml
```diff
@@ -34,3 +34,8 @@ jobs:
         run: |
           export PATH="/home/runner/.config/.foundry/bin:$PATH";
           make test;
+
+      - name: Check storage layout
+        run: |
+          export PATH="/home/runner/.config/.foundry/bin:$PATH";
+          make check-layout;
```
rjan90, line 43: This new step depends on the PR base ref, but the workflow doesn’t fetch it, so the check can silently degrade in shallow clones. Please either set `fetch-depth: 0` on checkout or explicitly fetch `${{ github.base_ref }}` before `make check-layout`.

rjan90 (changes requested):

rjan90:
@Chaitu-Tatipamula I pushed a small follow-up to tighten the storage-layout guardrail in https://github.com/FilOzone/pdp/pull/263/commits/38f3a764b62d2758d55d93478b03a52a99190993.

The previous snapshot generation was flattening `forge inspect` output down to top-level `label/slot/offset/type-label`, which meant nested layout changes inside packed structs could be invisible to CI. In practice, a destructive change inside `FeeStatus` or `PlannedUpgrade` could keep the same top-level type label and still pass `check_storage_layout.sh`.

This change updates the generated `PDPVerifierLayout.json` to preserve canonical nested type metadata, and updates the checker to compare the full canonicalized `type` JSON instead of a flattened string. That way member-level struct layout changes are treated as destructive and fail the check as intended.

rjan90 (approved):

Chaitu-Tatipamula:
> @Chaitu-Tatipamula I pushed a small follow-up to tighten the storage-layout guardrail in [38f3a76](https://github.com/FilOzone/pdp/commit/38f3a764b62d2758d55d93478b03a52a99190993).
> 
> The previous snapshot generation was flattening `forge inspect` output down to top-level `label/slot/offset/type-label`, which meant nested layout changes inside packed structs could be invisible to CI. In practice, a destructive change inside `FeeStatus` or `PlannedUpgrade` could keep the same top-level type label and still pass `check_storage_layout.sh`.
> 
> This change updates the generated `PDPVerifierLayout.json` to preserve canonical nested type metadata, and updates the checker to compare the full canonicalized `type` JSON instead of a flattened string. That way member-level struct layout changes are treated as destructive and fail the check as intended.

cool Thanks @rjan90

PR #266, opened 2026-05-05
feat(cid): raw size helpers in Cid library

rvagg:
Ref: https://github.com/FilOzone/filecoin-services/issues/451

### DIFF
--- a/src/Cids.sol
+++ b/src/Cids.sol
@@ -69,10 +69,12 @@ library Cids {
         return (1 << (uint256(height) + 5)) - (128 * padding) / 127;
     }
 
-    // rawPieceSize returns the raw (pre-Fr32-expansion) byte size from a PieceCIDv2's padding
-    // and height. Exact. Tree capacity is 2^(height+5) Fr32 bytes = 2^(height-2) * 127 raw bytes;
-    // padding is expressed in raw bytes. Smallest representable piece is height=2 (127 raw bytes).
-    // Reverts on height < 2 or padding exceeding capacity (Solidity 0.8 underflow check).
+    // rawPieceSize returns the exact raw (pre-Fr32-expansion) byte size of the data behind a
+    // PieceCIDv2. Per FRC-0069, the CID describes a binary tree of 2^height 32-byte leaves
+    // holding Fr32-expanded data (128 Fr32 bytes per 127 raw), and `padding` is the trailing
+    // zero bytes appended to the raw data before that expansion.
+    //   raw size = 2^height * 32 * 127/128 - padding = 2^(height-2) * 127 - padding
+    // Reverts on height < 2 or padding too large for the tree.
     function rawPieceSize(uint256 padding, uint8 height) internal pure returns (uint256) {
         return (1 << (uint256(height) - 2)) * 127 - padding;
     }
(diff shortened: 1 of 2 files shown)


rvagg:
This doesn't require an update of PDPVerifier but it will require a submodule update in filecoin-services so we could consider tagging it for that purpose alone without a deploy, or just submodule to an untagged commit, which would also be fine but slightly less clean.

src/Cids.sol
```diff
@@ -82,7 +90,14 @@ library Cids {
         return (1 << uint256(height)) - paddingLeafs;
     }
 
-    // Creates a CommPv2 CID from a raw size and hash digest according to FRC-0069.
+    // leafCountToRawSize gives an upper bound on raw bytes from a leaf count. Use rawPieceSize
+    // when padding and height are available; this is for aggregate counts (e.g. data set total).
+    // Overestimates by up to ~31 bytes per piece because per-piece leaf counts round up to 32.
```
Kubuxu: It can overestimate by up to 126 in case of a 1-byte piece (height=2, padding=126), it would have 4 leaves.

Kubuxu (approved):
LGTM apart from the comment nit

src/Cids.sol
```diff
@@ -69,6 +69,14 @@ library Cids {
         return (1 << (uint256(height) + 5)) - (128 * padding) / 127;
     }
 
+    // rawPieceSize returns the raw (pre-Fr32-expansion) byte size from a PieceCIDv2's padding
+    // and height. Exact. Tree capacity is 2^(height+5) Fr32 bytes = 2^(height-2) * 127 raw bytes;
```
ZenGround0: nit: I think a slightly clearer description is "Tree capacity is 2^(height) Fr32 bytes = 2^(height) * 2^5 bytes.  Total raw bytes is (127/128) * Tree capactiy - padding = 2^(height) * 2^5 * 2^-7 *127 - padding =  2^(height - 2)*127 - padding "  this may be annoyingly verbose so just a suggestion

ZenGround0 (approved):

src/Cids.sol
rvagg: Adding a 1-byte piece (which we've not enabled in the stack yet for various reasons, though PDPVerifier supports it): `addOnePiece` runs `Cids.leafCount(126, 2)` = `4 - ((128*126)/127 >> 5) = 4 - 3` = `1`, so the contract records 1 data-bearing leaf. `leafCountToRawSize(1)` = `(1*32*127)/128` = `31` against an actual 1 byte gives overestimate 30.

`rawPieceSize` stays exact (`1*127 - 126 = 1`) because FRC-0069 keeps the padding count in the CID, even though the data itself is zero-padded up to 127 for hashing.

I guess the 126 reading would land if you fed `2^height = 4` (the merkle tree's leaf count) in instead, but the contract has no tree-level accounting, only `Cids.leafCount`. We don't expect anyone to have reason to call with a notion of a tree. So per-piece overestimate is still bounded at 31 bytes, not 126.

Unless you're seeing something I'm missing here?

src/Cids.sol
rvagg: 👍 I found an in-between: describe it more clearly in prose and add the terse algorithmic form.

src/Cids.sol
rvagg: updated the comment for a little more clarity

Kubuxu: Ahhh, I confused the leafcount of the CommP tree and leafcount stored in PDP/FWSS

PR #270, opened 2026-05-19
feat: cleanup deposits

### LINKED ISSUE #269: [pdp] Simplified DataSet terminations and cleanup for GA
https://www.notion.so/filecoindev/Data-Set-Terminations-Clean-up-360dc41950c1801ebf0aff017a322e7f

Notes:
- Work is happening in https://github.com/FilOzone/pdp/tree/cleanup-pieces

wjmelements:
Reviewer @rvagg
Closes https://github.com/FilOzone/pdp/issues/269
[Spec](https://www.notion.so/filecoindev/Data-Set-Terminations-Clean-up-360dc41950c1801ebf0aff017a322e7f)
It is important that cleanup also works for existing datasets.
#### Changes
* implement cleanup deposit
* remove previous sybil enforcement
* test


src/PDPVerifier.sol
```diff
@@ -45,13 +38,18 @@ interface PDPListener {
 }
 
 uint256 constant NEW_DATA_SET_SENTINEL = 0;
+// Defaults
+uint256 constant NO_CHALLENGE_SCHEDULED = 0;
+uint256 constant NO_PROVEN_EPOCH = 0;
```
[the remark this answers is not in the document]
wjmelements: not for these

test/CleanupPieces.t.sol
```diff
@@ -0,0 +1,290 @@
...
+
+        for (uint256 pieceId = 0; pieceId < numPieces; pieceId++) {
+            // pieceCids stores a `bytes` value; for a 36-byte CID the header holds length*2+1=73.
+            // After delete, the header must be zero.
+            assertEq(
+                vm.load(address(pdpVerifier), keccak256(abi.encode(pieceId, cidRoot))),
+                bytes32(0),
+                "pieceCids header not cleared"
+            );
+            assertEq(
```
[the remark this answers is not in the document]
wjmelements: 168740f

src/PDPVerifier.sol
```diff
@@ -600,39 +556,124 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
...
+        }
+
+        uint256 pieceCount = nextPieceId[setId];
+        uint256 toClean = pieceCount < maxPieces ? pieceCount : maxPieces;
+
+        for (uint256 i = 0; i < toClean; i++) {
+            uint256 pieceId = pieceCount - 1 - i;
+            delete pieceCids[setId][pieceId];
+            delete pieceLeafCounts[setId][pieceId];
+            delete sumTreeCounts[setId][pieceId];
```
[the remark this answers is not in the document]
wjmelements: I think it does clear those slots, based on the test.

test/Fees.t.sol
```diff
@@ -40,9 +40,4 @@ contract PDPFeesTest is Test {
         uint256 expectedFee = 0.00023 ether; // 0.00023 FIL in attoFIL
         assertEq(PDPFees.DEFAULT_FEE_PER_TIB, expectedFee, "DEFAULT_FEE_PER_TIB should be 0.00023 FIL");
     }
```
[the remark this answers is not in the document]
wjmelements: 1 ether isn't the expected deposit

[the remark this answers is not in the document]
wjmelements: 5b77f601f89af1de65980eec7415a4e500c16f81

src/PDPVerifier.sol
```diff
@@ -174,21 +164,17 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
 
     PlannedUpgrade public nextUpgrade;
 
+    // FIL deposit collected at createDataSet and returned to whoever finalizes cleanup for that data set.
+    mapping(uint256 => uint256) cleanupDeposit;
+    // Block number when deleteDataSet was called, used to gate permissionless cleanupPieces calls.
+    mapping(uint256 => uint256) cleanupModeEpoch;
```
rvagg, line 180: should we just make this a single struct so we can easily extend it later if we need to (for incremental payout for example)?

src/PDPVerifier.sol
rvagg, line 1: `nextProvingPeriod` needs the check to reject setting `CLEANUP_MODE_SENTINEL` as the next proving period otherwise it can be set there by the SP

src/PDPVerifier.sol
```diff
@@ -600,39 +556,124 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
...
 
-        require(storageProvider[setId] == msg.sender, "Only the storage provider can delete data sets");
+        address sp = storageProvider[setId];
+        require(sp != address(0), DataSetNotLive());
+        require(nextChallengeEpoch[setId] != CLEANUP_MODE_SENTINEL, DataSetAlreadyInCleanup());
+
+        // Permissionless if the SP has been inactive for more than INACTIVITY_WINDOW blocks.
+        if (block.number <= dataSetLastProvenEpoch[setId] + INACTIVITY_WINDOW) {
+            require(msg.sender == sp, OnlyStorageProviderCanDelete());
+        }
```
rvagg, line 606: We need to solve for the data-set-never-proven case where `dataSetLastProvenEpoch[setId]` remains `0` so becomes immediately eligible for deletion on creation; I guess your existing tests pass because they start `block.number` at `0`.
We don't have a data set creation epoch recorded, we could use that if we had one, maybe we should add one? The other option would be to set `dataSetLastProvenEpoch[setId]` to be `block.number` on creation but (a) that makes weird state inspection ("when was it last proven?" doesn't give a reliable response) and (b) what about existing data sets that have never been proven--if this gets deployed tomorrow and there was a data set created today it may have `0` here and be immediately eligible.

src/PDPVerifier.sol
rvagg: also should we make these public? could make it interesting for dashboards or keepers that want to work on this

wjmelements: I would like for all of the data set fields to be in a data set struct but this project went in the opposite direction.

src/PDPVerifier.sol
wjmelements: The best way to do this would be to enforce a limit on how far in the future the next proving period can be.

src/PDPVerifier.sol
wjmelements: Dashboards and keepers don't need public view methods. Public view methods are for on-chain access. I think they should be checked offchain.

src/PDPVerifier.sol
wjmelements: 5cdb655cc94e866678039e70c5cfb1105bd07ff0

src/PDPVerifier.sol
```diff
@@ -109,14 +107,20 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
...
     // PoRep precommit and PoRep provecommit phases.
     //
     // We keep this around for future portability to a variety of environments with different assumptions
     // behind their challenge randomness sampling methods.
-    uint256 challengeFinality;
+
+    /// @custom:oz-upgrades-unsafe-allow state-variable-immutable
+    uint256 private immutable CHALLENGE_FINALITY;
+    /// @custom:oz-upgrades-unsafe-allow state-variable-immutable
+    uint256 private constant MAX_FINALITY = 259200; // 90 days of 30-second epochs
```
wjmelements: perhaps this number needs to be smaller, accounting for inactivity window

src/PDPVerifier.sol
wjmelements: An empty data set can be proven vacuously.

We can workaround for existing data sets if we forbid permissionless deletion until a certain number of epochs after the ugprade.

There is another case to consider: if the last piece is removed, we reset the last proven and next challenge epochs (`DataSetEmpty`). I plan to change this behavior.

src/PDPVerifier.sol
wjmelements: d3590ac7bd54c092be3c66cc345aa0af45d51966

src/PDPVerifier.sol
wjmelements: aa7e50d9b7b8682e958755ee43f0336e1b981fa4

tools/deploy-mainnet.sh
```diff
@@ -39,7 +39,7 @@ if [ "$USDFC_TOKEN_ADDRESS" = "$ZERO_ADDRESS" ] || [ "$PAYMENTS_CONTRACT_ADDRESS
 fi
 
 NONCE="$(cast nonce --rpc-url "$RPC_URL" "$ADDR")"
-VERIFIER_IMPLEMENTATION_ADDRESS=$(forge create --rpc-url "$RPC_URL" --keystore "$KEYSTORE" --password "$PASSWORD" --broadcast  --nonce $NONCE --chain-id 314  src/PDPVerifier.sol:PDPVerifier --constructor-args "$VERIFIER_INIT_COUNTER" "$USDFC_TOKEN_ADDRESS" "$USDFC_SYBIL_FEE" "$PAYMENTS_CONTRACT_ADDRESS" | grep "Deployed to" | awk '{print $3}')
+VERIFIER_IMPLEMENTATION_ADDRESS=$(forge create --rpc-url "$RPC_URL" --keystore "$KEYSTORE" --password "$PASSWORD" --broadcast  --nonce $NONCE --chain-id 314  src/PDPVerifier.sol:PDPVerifier --constructor-args "$VERIFIER_INIT_COUNTER" "$CHALLENGE_FINALITY" "$USDFC_TOKEN_ADDRESS" "$USDFC_SYBIL_FEE" "$PAYMENTS_CONTRACT_ADDRESS" | grep "Deployed to" | awk '{print $3}')
```
rvagg: the constructor args are out of sync now, deploy-mainnet.sh, deploy-calibnet.sh and upgrade-contract.sh all have 5 args, we need to remove the usdfc args

src/PDPVerifier.sol
```diff
@@ -946,16 +1011,18 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
 
         // Bring added pieces into proving set
         challengeRange[setId] = dataSetLeafCount[setId];
-        if (challengeEpoch < block.number + challengeFinality) {
-            revert("challenge epoch must be at least challengeFinality epochs in the future");
+        {
+            uint256 delayEpochs = challengeEpoch - block.number;
```
rvagg, line 1015: by switching the `if` to a subtraction you've got a potential underflow here, if `challengeEpoch` is less than `block.number` then we get weirdness, I think a panic?

```suggestion
            uint256 delayEpochs = challengeEpoch > block.number ? challengeEpoch - block.number : 0;
```

that should give us the `InsufficientChallengeDelay` which would be expected for in-the-past epoch.

src/PDPVerifier.sol
```diff
@@ -174,32 +174,30 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
...
 
     /// @custom:oz-upgrades-unsafe-allow constructor
-    constructor(
-        uint64 _initializerVersion,
-        address _usdfcTokenAddress,
-        uint256 _usdfcSybilFee,
-        address _paymentsContractAddress
-    ) {
+    constructor(uint64 _initializerVersion, uint256 _challengeFinality) {
+        require(_challengeFinality < MAX_FINALITY / 10);
```
rvagg: I think a msg would help here for a deployer doing something unexpected?

rvagg (approved):
approving in expectation of resolution of the bash script fixups, and an eye to the underflow panic vs proper error

otherwise lgtm!

Kubuxu (approved):
Looks good
One nit that is probably not worth fixing:
`dataSetLastProvenEpoch` became something else, something closer to `lastModified`.
Maybe a comment on its actual uses would be useful.

src/PDPVerifier.sol
```diff
@@ -228,60 +226,32 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
...
-            }
-        }
-    }
-
-    function _getPaymentsUsdfcBalance() internal view returns (uint256) {
-        if (PAYMENTS_CONTRACT_ADDRESS == address(0) || USDFC_TOKEN_ADDRESS == address(0)) return 0;
-        (uint256 funds,,,) =
-            IFilecoinPay(PAYMENTS_CONTRACT_ADDRESS).accounts(USDFC_TOKEN_ADDRESS, PAYMENTS_CONTRACT_ADDRESS);
-        return funds;
-    }
```
ZenGround0, line 270: Since we are removing sybil fee paid by clients in favor of posting sybil bond paid by SP I want to make sure that we are not walking right back into the same problems that we solved for SP.  It MUST be the case that SPs can quickly reclaim their bond if they are being DOSed by clients.  If there is any chance that the funds can't be recovered then this is not an acceptable situation.

wjmelements: We aren't removing the sybil fee from FWSS, but we are removing this check.

src/PDPVerifier.sol
wjmelements: It is ok to Panic from underflow.

tools/deploy-mainnet.sh
wjmelements: ec6f8aaeaa5640aa37a42e56e0544373154efa4e

src/PDPVerifier.sol
wjmelements: If we let it underflow it would hit the other error and I think that could be fine.

wjmelements: 9d430137f9215da2cd0f2059e9a6236238a51bdb

src/PDPVerifier.sol
wjmelements: 4b031fb61a5d4040aa22091e8ac92c65e533688e

src/PDPVerifier.sol
wjmelements: It is called the "Network Congestion Fee" [here](https://github.com/FilOzone/filecoin-services/issues/469#issue-4363823781)

PR #272, opened 2026-05-22
docs(changelog): draft v3.4.0 release notes and version bump

rjan90:
## Summary

- Draft `v3.4.0` release notes in `CHANGELOG.md` with date and deployment fields left as `TBD` until rollout execution.
- Bump `PDPVerifier.VERSION` to `3.4.0` so the eventual implementation bytecode reports the intended release version.

## Notes

- This intentionally excludes the local `pdpverifier-upgrade.md` process/template edits; those are still being iterated separately.
- Deployment addresses, verification links, and release date should be finalized after Calibration and Mainnet rollout execution.

## Validation

- `forge test` (`179` tests passing)

### DIFF
--- a/src/PDPVerifier.sol
+++ b/src/PDPVerifier.sol
@@ -197,7 +197,7 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         feeStatus.nextFeePerTiB = PDPFees.DEFAULT_FEE_PER_TIB;
     }
 
-    string public constant VERSION = "3.3.0";
+    string public constant VERSION = "3.4.0";
 
     event ContractUpgraded(string version, address implementation);
     event UpgradeAnnounced(PlannedUpgrade plannedUpgrade);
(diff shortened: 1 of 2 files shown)


src/PDPVerifier.sol
```diff
@@ -197,7 +197,7 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
         feeStatus.nextFeePerTiB = PDPFees.DEFAULT_FEE_PER_TIB;
     }
 
-    string public constant VERSION = "3.3.0";
```
wjmelements, line 200: The previous change (#270) bumped it to 3.3.0. I don't have a strong opinion on this though. But we didn't deploy any 3.3.0.

rjan90: The reason for the fast-forward here is that we shipped v3.3.0 release https://github.com/FilOzone/pdp/releases/tag/v3.3.0 which only had the library changes.

So the fast-forward from v3.2.0 to v3.4.0 in the deployed bytecode is to align them

rjan90: Added a changelog note explaining the version fast-forward: the live proxies still report `3.2.0`, `v3.3.0` was library-only with no PDPVerifier deployment, and this rollout intentionally moves deployed `VERSION()` from `3.2.0` to `3.4.0`.

Kubuxu (approved):

PR #274, opened 2026-05-26
feat: cleanup mode view

wjmelements:
Reviewer @zenground0
CC @LexLuthr @rvagg
#### Changes
* allow view for getNextPieceId and getNextChallengeEpoch while in cleanup mode
* replace string error with abi error
* remove redundant setId checks

### DIFF
--- a/src/PDPVerifier.sol
+++ b/src/PDPVerifier.sol
@@ -250,7 +250,7 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
 
     // Returns false if the data set is 1) not yet created 2) fully deleted (cleanup complete)
     function dataSetExists(uint256 setId) internal view returns (bool) {
-        return setId < nextDataSetId && storageProvider[setId] != address(0);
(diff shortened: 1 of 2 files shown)


src/PDPVerifier.sol
```diff
@@ -248,10 +248,14 @@ contract PDPVerifier is Initializable, UUPSUpgradeable, OwnableUpgradeable {
...
     }
 
+    // Returns false if the data set is 1) not yet created 2) fully deleted (cleanup complete)
+    function dataSetExists(uint256 setId) internal view returns (bool) {
+        return storageProvider[setId] != address(0);
+    }
+
     // Returns false if the data set is 1) not yet created 2) deleted or in cleanup mode
     function dataSetLive(uint256 setId) public view returns (bool) {
-        return setId < nextDataSetId && storageProvider[setId] != address(0)
```
wjmelements, line 253: we don't have to check setId < nextDataSetId because it is impossible for storageProvider[setId] to be set otherwise

ZenGround0 (approved):
LGTM

PR #279, opened 2026-06-05
fix(cleanup): anchor cleanupPieces gate to last proving activity

rvagg:
The permissionless gate was anchored to cleanup-mode entry, so the abandonment path took two INACTIVITY_WINDOWs (~60 days): a third party could delete an abandoned data set after 30 days but had to wait another 30 to clean up and collect the deposit. Both gates now share _withinActivityWindow, making delete and cleanup permissionless at the same moment. Deprecates the now-unused cleanupModeEpoch in place.

---

`deleteDataSet` becomes permissionless after `INACTIVITY_WINDOW` (~30 days) without proving activity, but cleanupPieces restarted that clock from the delete itself. The abandonment path therefore took ~60 days, and the cleanup deposit, meant as the bounty for whoever tears down an abandoned data set, was unclaimable by the deleter for a month and then up for grabs by anyone, breaking the incentive to do the (expensive) permissionless delete at all.

Fix: both functions now share one gate, `_withinActivityWindow()`, anchored to dataSetLastProvenEpoch (legacy fallback to `LEGACY_ACTIVITY_EPOCH`). That epoch is frozen once a data set enters cleanup mode (`provePossession`/`nextProvingPeriod`/`addPieces` all revert on deleted data sets), so a data set abandoned enough to delete is immediately cleanable by the same caller, while an actively-proving SP keeps its ~30-day exclusive cleanup window after a voluntary delete. The call ordering (delete, then cleanup) is unchanged: it's enforced by the `CLEANUP_MODE_SENTINEL`, which never depended on the removed anchor. `cleanupModeEpoch` is deprecated in place; `_finalizeCleanup` still clears v3.4.0 residue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 
+### Fixed
+- `cleanupPieces()` now uses the same permission gate as `deleteDataSet()`, anchored to last proving activity instead of cleanup-mode entry. The abandonment path previously required two full `INACTIVITY_WINDOW` periods (~60 days); a permissionless deleter can now clean up and collect the deposit immediately. An SP deleting after exceeding the inactivity window no longer gets an exclusive cleanup period (the data set was already permissionlessly deletable). The unused `cleanupModeEpoch` storage is deprecated in place.
+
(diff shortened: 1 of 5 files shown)


wjmelements (approved):

wjmelements:
Nice. Too bad we already deployed the `cleanupModeEpoch`.

